PR #9, opened 2017-07-28
Base Architecture

nevir:
Herein lies an overview of the architecture as well as the interfaces (and more detailed algorithms) that satisfy it.

I'd suggest starting with [`docs/Architecture.md`](https://github.com/convoyinc/apollo-cache-hermes/pull/9/files?short_path=ca9b7cf#diff-ca9b7cf3563e866c893889395228da20), which explains the broad strokes.  Then, follow up with the individual interfaces; roughly in this order:

`src/*Snapshot.ts` which describe the shape of the various pieces of the cache.

`src/Cache.ts` which is the entry point into this module, to (roughly) satisfy [Apollo Client 2.0](https://github.com/apollographql/apollo-client/pull/1941)'s [new cache API](https://github.com/apollographql/apollo-client/pull/1921)

`src/operations/*` for the actual work we perform against the cache.

---

Note that I haven't gotten to the following pieces of the architecture:

* Query observers.
* Apollo Client's abstraction around transactions, and optimistic transactions.

However, I'm pretty confident they can be built given all the components outlined here.  Will be fast-following this PR with their architecture, as well.


docs/Architecture.md
```diff
@@ -0,0 +1,123 @@
...
+
+4. When querying the cache for values, it should prefer values present in optimistic updates over those in the underlying cache.
+
+Due to (2) and (3), we know that we cannot blindly merge optimistic updates into an existing snapshot - and that we must track the base cache snapshot.  Also, due to (3), whenever we receive new values from the server, we effectively need to update the raw cache snapshot, and then replay optimistic updates.
+
+The approach that seems best here is to:
+
+* Track all optimistic updates individually via an [optimistic state queue](../src/OptimisticUpdateQueue.ts), where each update is represented in the same format as a GraphQL response payload.
+
+* The cache tracks both a base graph snapshot and - if there are active optimistic updates - an optimistic graph snapshot.  Every time either the raw snapshot changes, or the optimistic state queue changes, we regenerate the unified snapshot by replaying the optimistic updates on top of the base snapshot.
```
EvanBoyle: I think that it would be helpful to reason about optimistic updates by calling out explicit scenarios:

1. User is offline for an extended period of time. App should behave as normal and network calls can be replayed once reconnected. 
2. User is mostly connected, updates will succeed the vast majority of the time. 

In case 1 it seems reasonable to merge state changes directly to the cache. During an extended offline session, these changes would likely pile up and replaying events could become really slow. 

Case 2 (well connected) seems like fewer updates would pile up at a given point in time and replaying them each time server data comes in wouldn't be too big of an issue. 

One idea for a simplified optimistic update mechanism could be to merge all optimistic updates to the cache while still maintaining a queue of state changes. If one of the updates fails, you could bust the cache pulling fresh state from the server and then replay the additional queued state changes.

nevir: > In case 1 it seems reasonable to merge state changes directly to the cache. During an extended offline session, these changes would likely pile up and replaying events could become really slow.

Unfortunately, I don't think it's safe to merge them directly into the cache:

* When we do end up getting back online, after the update finally flushes, the server may disagree, but we still want to represent the optimistic state until _all_ optimistic updates have flushed.  

* Those requests could still fail at some point, which should invalidate that specific update, but not others.  If we merge directly into the cache, we lose all ability to safely roll back.

---

One strategy to address growing sets of updates is to merge all the updates into one delta (as opposed to replaying one at a time), and just apply that when the base store changes.  It gets a little tricky, though, as each update can be rooted at a different node in the graph

docs/Architecture.md
```diff
@@ -0,0 +1,123 @@
...
+
+Because the cache is built to store values in a format that can be directly returned (for un-parameterized edges), most of the work to perform a query revolves around making sure that the cache can satisfy the query.  The high level approach to performing a query is roughly:
+
+1. Pre-process the query (if not already done), extracting paths to parameterized edges.
+
+2. If there are parameterized edges in the query, fill in the object path up to them, taking advantage of object prototypes to point to the underlying nodes.
+
+3. Verify that the query is satisfied by the cache.  _The naive approach is to walk the selection set(s) expressed by the query; it's probably good enough for now_.
+
+4. Return the query root, or view on top of it via (3).
```
RylanH: Do you mean (2)?

docs/Architecture.md
```diff
@@ -0,0 +1,123 @@
...
+
+Generally, when reading, we want to return whatever data we have, as well as a status indicating whether the query was completely satisfying.  The caller can determine what to do if not satisfied.
```
RylanH: satisfying -> satisified?

docs/Architecture.md
nevir: Doh, yup.

docs/Architecture.md
nevir: We want our queries to feel like:

taqcv0s

PR #26, opened 2017-08-09
Make things work with apollo 2.0

vincecoppola:
Small changes to make things easier to plug in to apollo 2.0.

### DIFF
--- a/src/Cache.ts
+++ b/src/Cache.ts
@@ -26,17 +26,20 @@ export const defaultConfiguration:Configuration = {
  */
 export class Cache implements Queryable {
 
+  /** The cache-wide configuration. */
+  private _config: Configuration;
+
   /** The current version of the cache. */
   private _snapshot: CacheSnapshot;
 
   /** All active query observers. */
   private _observers: QueryObserver[] = [];
 
-  constructor(private readonly _config: Configuration) {
+  constructor(config?: Configuration) {
     const initialGraphSnapshot = new GraphSnapshot();
     this._snapshot = new CacheSnapshot(initialGraphSnapshot, initialGraphSnapshot, new OptimisticUpdateQueue());
 
-    this._config = lodashDefaults(_config, defaultConfiguration);
+    this._config = lodashDefaults(config, defaultConfiguration);
   }
 
   /**
--- a/src/Cache.ts
+++ b/src/Cache.ts
@@ -7,7 +7,7 @@ import { CacheSnapshot } from './CacheSnapshot';
 import { Configuration } from './Configuration';
 import { GraphSnapshot } from './GraphSnapshot';
 import { QueryObserver, read } from './operations';
-import { OptimisticUpdateQueue } from './OptimisticUpdateQueue';
+import { OptimisticUpdate, OptimisticUpdateQueue } from './OptimisticUpdateQueue';
 import { ChangeId, NodeId, Query } from './schema';
 
 export type TransactionCallback = (transaction: CacheTransaction) => void;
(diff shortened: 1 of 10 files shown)


src/Cache.ts
```diff
@@ -1,3 +1,5 @@
+import { defaults, get } from 'lodash';
```
nevir: nit: let's use the specific `lodash` packages (e.g. `lodash.get`, `lodash.defaults`) to reduce our footprint

src/Cache.ts
```diff
@@ -9,6 +11,12 @@ import { ChangeId, NodeId, Query } from './schema';
 
 export type TransactionCallback = (transaction: CacheTransaction) => void;
 
+export const defaultConfiguration:Configuration = {
+  entityIdForNode(node) {
+    return get(node, 'id', undefined);
```
nevir: `undefined` is already the default for this guy

src/Cache.ts
```diff
@@ -17,14 +25,19 @@ export type TransactionCallback = (transaction: CacheTransaction) => void;
...
 export class Cache implements Queryable {
 
-  /** Configuration used by various operations made against the cache. */
-  private readonly _config: Configuration;
-
   /** The current version of the cache. */
-  private _snapshot: CacheSnapshot;
+  private _snapshot: CacheSnapshot = new CacheSnapshot(
+    new GraphSnapshot,
+    new GraphSnapshot,
```
nevir: Ah, this should be the same object that's passed in for `baseline` (they should only differ when there are actual optimistic updates active)

src/Cache.ts
```diff
@@ -17,14 +25,19 @@ export type TransactionCallback = (transaction: CacheTransaction) => void;
...
+    new OptimisticUpdateQueue,
+  );
 
   /** All active query observers. */
-  private _observers: QueryObserver[];
+  private _observers: QueryObserver[] = [];
+
+  constructor(private readonly _config: Configuration) {
+    this._config = defaults(_config, defaultConfiguration);
```
nevir: 👍

src/schema.ts
```diff
@@ -21,9 +21,9 @@ export type EntityId = NodeId;
  * There are a few pre-defined nodes present in all schemas.
  */
 export enum StaticNodeId {
-  QueryRoot = '☣QueryRoot',
-  MutationRoot = '☣MutationRoot',
-  SubscriptionRoot = '☣SubscriptionRoot',
+  QueryRoot = 'ROOT_QUERY',
```
nevir: Does Apollo mandate these?

src/util/references.ts
```diff
@@ -1,4 +1,4 @@
-import lodashIsEqual = require('lodash.isequal');
+import lodashIsEqual = require('lodash/isequal');
```
nevir: 👎  for reasons above

src/Cache.ts
vincecoppola: Gotcha, so instantiate it once and then pass the single instance for both, yeah?

src/schema.ts
vincecoppola: It's hard coded in their code, at least. This was what tripped me up for the longest amount of time.

src/util/references.ts
vincecoppola: This was throwing for me but maybe I messed something up with my packages

src/schema.ts
nevir: 👍

src/Cache.ts
nevir: Yup!

PR #71, opened 2017-08-21
Rename thing to make add alias easier

yuit:


src/CacheTransaction.ts
```diff
@@ -16,10 +16,10 @@ import { addToSet } from './util';
 export class CacheTransaction implements Queryable {
 
   /** The set of nodes edited throughout the transaction. */
-  private _editedNodeIds = new Set() as Set<NodeId>;
+  private _editedNodeIds = new Set<NodeId>();
```
nevir: 👎 I'd like to stick with `as T` style assertions so that we're consistent w/ our other code bases (e.g. because of `.tsx` everywhere)

src/NodeSnapshot.ts
```diff
@@ -7,7 +7,7 @@ import { NodeId } from './schema';
 export class NodeSnapshot {
   constructor(
     /** A reference to the node this snapshot is about. */
-    readonly node?: any,
+    public node?: object,
```
nevir: referenced node values aren't guaranteed to be objects - could be scalar too

src/NodeSnapshot.ts
```diff
@@ -7,7 +7,7 @@ import { NodeId } from './schema';
...
     /** Other node snapshots that point to this one. */
     readonly inbound?: NodeSnapshot.Reference[],
```
nevir: Hmm, let's make them all writable properties while we're at it

src/util/ast.ts
```diff
@@ -64,22 +69,31 @@ export class VariableArgument {
 }
 
 /**
+ * A recursive map where the keys indicate the path to any edge in a result set
+ * that contain a parameterized edge.
+ */
+export interface DynamicEdgeMap {
```
nevir: 👍 much better name

src/operations/SnapshotEditor.ts
```diff
@@ -398,14 +400,14 @@ export class SnapshotEditor {
...
-    const snapshot = this.getSnapshot(id);
+    const snapshot = this.getNodeSnapshot(id);
     return snapshot ? snapshot.node : undefined;
   }
 
   /**
    * Retrieve the _latest_ version of a node snapshot.
    */
-  private getSnapshot(id: NodeId) {
+  private getNodeSnapshot(id: NodeId) {
```
nevir: 👍

src/operations/SnapshotEditor.ts
```diff
@@ -128,21 +128,22 @@ export class SnapshotEditor {
...
-    const visitedNodes = new Set<any>();
+    const visitedNodes = new Set<object>();
 
     while (queue.length) {
-      const { containerId, containerPayload, visitRoot, edges } = queue.pop() as MergeQueueItem;
-      const containerSnapshot = this.getSnapshot(containerId);
-      const container = containerSnapshot ? containerSnapshot.node : undefined;
+      const { containerId, containerPayload, visitRoot, edges } = queue.pop()!;
+      const containerSnapshot = this.getNodeSnapshot(containerId);
+      const container = this.get(containerId);
```
nevir: IMO we want the previous version - `.get` is just going to call `getNodeSnapshot` again under the covers

src/operations/SnapshotEditor.ts
```diff
@@ -221,6 +222,7 @@ export class SnapshotEditor {
           // The payload is now referencing a new entity.  We want to update it,
           // but not until we've updated the values of our entities first.
           if (prevNodeId !== nextNodeId) {
+            // We have spread "path" so that we pass in new array.
```
nevir: Ah sorry - real reason is that `path` is mutated by `walkPayload`; this ensures that we get a static copy of it

src/util/ast.ts
```diff
@@ -13,6 +13,19 @@ import { // eslint-disable-line import/no-extraneous-dependencies, import/no-unr
...
+ * NOTE: this list is NOT complete and only include node _kind_ we use
+ */
+export enum GraphqlNodeKind {
+  Field = 'Field',
+  FragmentSpread = 'FragmentSpread',
+  FragmentDefinition = 'FragmentDefinition',
+  InlineFragment = 'InlineFragment',
+  Name = 'Name',
+  OperationDefinition = 'OperationDefinition',
+}
```
nevir: I always felt a bit weird about enums like this:

* The name and value of the enum will _always_ match (what's the benefit over a string literal?)
* We're likely to forget to clean these up when we stop using a node type
* (You didn't do it for the [value types](https://github.com/convoyinc/apollo-cache-hermes/pull/71/files#diff-f1c479fe7715b995669798dca275b392R172) :P)

(I don't feel strongly either way, just curious what your take is)

src/util/ast.ts
```diff
@@ -64,22 +69,31 @@ export class VariableArgument {
...
-export interface EdgeArgumentObject { [Key: string]: EdgeArgument }
-export type EdgeArgument = EdgeArgumentScalar | EdgeArgumentArray | EdgeArgumentObject;
+export interface EdgeArgumentArray extends Array<ParameterizedEdgeArgumentValue> {}
+export interface ParameterizedEdgeArguments { [argumentName: string]: ParameterizedEdgeArgumentValue }
+export type ParameterizedEdgeArgumentValue = EdgeArgumentScalar | EdgeArgumentArray | ParameterizedEdgeArguments;
 
 /**
- * Represents a parameterized edge (within an edge map).
+ * Represent dynamic information: alias, parameterized arguments,
+ * directives (if existed) of NodeSnapshot in GraphSnapshot.
```
nevir: style nit: wrap comments at 80 characters; rest of the code can be ¯\\\_(ツ)\_/¯

src/util/ast.ts
```diff
@@ -151,35 +166,37 @@ function _buildParameterizedEdgeArgs(field: FieldNode & { arguments: ArgumentNod
...
     return node.values.map(_valueFromNode);
-  } else if (node.kind === 'ObjectValue') {
+  case 'ObjectValue': {
     const value = {};
     for (const field of node.fields) {
       value[field.name.value] = _valueFromNode(field.value);
     }
     return value;
-  } else {
+  }
```
nevir: style nit: drop the `}`'s, they're optional here, and introduce an extra lexical scope

src/util/ast.ts
```diff
@@ -151,35 +166,37 @@ function _buildParameterizedEdgeArgs(field: FieldNode & { arguments: ArgumentNod
...
+  default:
     return node.value;
   }
 }
 
 /**
  * Sub values in for any variables required by an edge's args.
  */
-export function expandEdgeArguments(edge: ParameterizedEdge, variables: object = {}): object {
+export function expandEdgeArguments(edge: DynamicEdge, variables: object = {}): object {
```
nevir: Let's enforce the contract of this method with:

```ts
type DynamicEdgeWithArgs = DynamicEdge & { parameterizedEdgeArgs: ParameterizedEdgeArguments };

…

export function expandEdgeArguments(edge: DynamicEdgeWithArgs, variables: object = {}): object {
```

src/util/ast.ts
```diff
@@ -244,10 +258,9 @@ export function addTypenameToDocument(doc: DocumentNode) {
   const docClone = lodashCloneDeep(doc);
 
   docClone.definitions.forEach((definition: DefinitionNode) => {
-    const isRoot = definition.kind === 'OperationDefinition';
     addTypenameToSelectionSet(
       (definition as OperationDefinitionNode).selectionSet,
-      isRoot,
+      /* isRoot*/ definition.kind === GraphqlNodeKind.OperationDefinition,
```
nevir: style nit: even spacing inside the comment!

src/util/tree.ts
```diff
@@ -84,7 +84,7 @@ export function walkPayload(
     }
 
     // Note that in all cases, we push nodes onto the stack in _reverse_ order,
-    // so that we visit nodes in iteration order (the stack is FIFO).
+    // so that we visit nodes in iteration order (the stack is LIFO).
```
nevir: Hah!  Very true

src/util/ast.ts
yuit: You bring up good point about cleaning up -> totally agree with that :) 
I like do it as string enum to minimize any typing error of the string (arguably it will be caught by the fact that type guard won't work) 

Also: note to self string enum won't type guard when using with switch statement : (

nevir: > I like do it as string enum to minimize any typing error of the string (arguably it will be caught by the fact that type guard won't work)

Yeah, and it'll be pretty clear by the inferred type within the guarded scope

src/CacheTransaction.ts
nevir: Wait, this is totally ok, ignore me

src/util/ast.ts
yuit: we have eslint rule that disallow declaring block scope variable in case clause :(

src/util/ast.ts
yuit: nice I love it 🍔

src/util/ast.ts
nevir: oh, oh!  I guess that's fair

PR #76, opened 2017-08-22
Refactor NodeSnapshot into multiple classes

nevir:
Starts to pave the way for `ParameterizedEntitySnapshot`s, and maybe `QuerySnapshot`s.


src/nodes/clone.ts
```diff
@@ -0,0 +1,32 @@
...
+    node = [...parent.node];
+  } else if (isObject(parent.node)) {
+    node = { ...parent.node };
+  } else {
+    node = parent.node;
+  }
+  const inbound = parent.inbound ? [...parent.inbound] : undefined;
+  const outbound = parent.outbound ? [...parent.outbound] : undefined;
+
+  const NodeSnapshotClass = Object.getPrototypeOf(parent).constructor;
```
yuit: Q: why can't we just use `instanceof`?

src/nodes/EntitySnapshot.ts
```diff
@@ -0,0 +1,16 @@
+import { NodeReference, NodeSnapshot } from './NodeSnapshot';
+
+/**
+ * Maintains a reference to a single entity within the cached graph, and any
+ * bookkeeping metadata associated with it.
+ */
+export class EntitySnapshot implements NodeSnapshot {
```
yuit: Nit: could you make comment that the `EnitySnapshot` is for static only

src/operations/SnapshotEditor.ts
```diff
@@ -430,29 +415,48 @@ export class SnapshotEditor {
    * Ensures that we have built a new version of a snapshot for node `id` (and
    * that it is referenced by `_newNodes`).
    */
-  private _ensureNewSnapshot(id: NodeId, initialValue?: object): NodeSnapshot {
-    let newSnapshot;
+  private _ensureNewSnapshot(id: NodeId): NodeSnapshot {
+    let parent;
```
yuit: nit; provide typing for the parent

src/operations/SnapshotEditor.ts
```diff
@@ -430,29 +415,48 @@ export class SnapshotEditor {
...
   }
 
+  /**
+   * Ensures that there is a ParameterizedValueSnapshot for the given field.
+   */
+  _ensureParameterizedValueSnapshot(containerId: NodeId, path: PathPart[], parameterizedEdge: ParameterizedEdge, variables: object) {
+    const edgeArguments = expandEdgeArguments(parameterizedEdge, variables);
+    const edgeId = nodeIdForParameterizedValue(containerId, path, edgeArguments);
+
+    const containerSnapshot = this.getSnapshot(containerId);
```
yuit: This is not that clear why we do `getSnapshot` and manually check it is is exist instead use `_ensureSnapshot`... wonder if it can be clearer

src/nodes/clone.ts
yuit: If need be we can introduce `kind` flag  😅

src/nodes/ParameterizedValueSnapshot.ts
```diff
@@ -0,0 +1,20 @@
...
+ * contained within some other node.
+ *
+ * These values are stored outside of the entity that contains them, as the
+ * entity node is reserved for static values.  At read time, these values are
+ * overlaid on top of the static values of the entity that contains them.
+ */
+export class ParameterizedValueSnapshot implements NodeSnapshot {
+  constructor(
+    /** A reference to the entity this snapshot is about. */
+    readonly node?: any,
```
yuit: not a big fan 😞  but I don't know how to better do this

src/operations/SnapshotEditor.ts
```diff
@@ -430,29 +415,48 @@ export class SnapshotEditor {
...
+    let parent;
     if (id in this._newNodes) {
       const current = this._newNodes[id];
       // We may have deleted the node.
       if (current) return current;
-      // If so, we should start fresh.
-      newSnapshot = new NodeSnapshot();
+      // We don't have enough information to safely recover in this case for
+      // non-entity nodes; let's not risk it.
+      throw new Error(`Re-creating deleted cache nodes mid-transaction is not yet supported`);
```
yuit: Q: how is this happen? since we do the garbage collecting last?

src/nodes/clone.ts
nevir: I was trying too hard to use `switch` there - 👍 much better with instanceof

src/operations/SnapshotEditor.ts
nevir: No need - TS properly infers it to `NodeSnapshot | undefined`

src/operations/SnapshotEditor.ts
nevir: good point!

src/operations/SnapshotEditor.ts
nevir: We don't want to `_ensureSnapshot` at the top level - if the parameterized value was already created in a prior version of the graph, we don't want to create a new copy of it (there may be no change)

src/nodes/ParameterizedValueSnapshot.ts
nevir: It literally can be `any` type :P

src/operations/SnapshotEditor.ts
yuit: You are in those camp who like type evolving thing

nevir: Hey, if the compiler can infer it - let it (except for public interfaces, which definitely should be locked in)

PR #79, opened 2017-08-22
Support for alias

yuit:
Should be merge after PR #71

### DIFF
--- a/src/GraphSnapshot.ts
+++ b/src/GraphSnapshot.ts
@@ -26,11 +26,7 @@ export class GraphSnapshot {
   /**
    * Retrieves the value identified by `id`.
    */
-<<<<<<< HEAD
   get(id: NodeId): Readonly<any> | undefined {
-=======
-  get(id: NodeId): any | undefined {
->>>>>>> Address PR: rename getGraphNodeSnapshot
     const snapshot = this.getNodeSnapshot(id);
     return snapshot ? snapshot.node : undefined;
   }
@@ -47,11 +43,7 @@ export class GraphSnapshot {
    *
    * @internal
    */
-<<<<<<< HEAD
   getNodeSnapshot(id: NodeId): Readonly<NodeSnapshot> | undefined {
-=======
-  getNodeSnapshot(id: NodeId): NodeSnapshot | undefined {
->>>>>>> Address PR: rename getGraphNodeSnapshot
     return this._values[id];
   }
 
--- a/src/GraphSnapshot.ts
+++ b/src/GraphSnapshot.ts
@@ -27,7 +27,7 @@ export class GraphSnapshot {
    * Retrieves the value identified by `id`.
    */
   get(id: NodeId): Readonly<any> | undefined {
-    const snapshot = this.getSnapshot(id);
+    const snapshot = this.getNodeSnapshot(id);
     return snapshot ? snapshot.node : undefined;
   }
 
@@ -43,7 +43,7 @@ export class GraphSnapshot {
    *
    * @internal
    */
-  getSnapshot(id: NodeId): Readonly<NodeSnapshot> | undefined {
+  getNodeSnapshot(id: NodeId): Readonly<NodeSnapshot> | undefined {
     return this._values[id];
   }
 
--- a/src/GraphSnapshot.ts
+++ b/src/GraphSnapshot.ts
@@ -26,7 +26,11 @@ export class GraphSnapshot {
   /**
    * Retrieves the value identified by `id`.
    */
+<<<<<<< HEAD
   get(id: NodeId): Readonly<any> | undefined {
+=======
+  get(id: NodeId): any | undefined {
+>>>>>>> Address PR: rename getGraphNodeSnapshot
     const snapshot = this.getNodeSnapshot(id);
     return snapshot ? snapshot.node : undefined;
   }
@@ -43,7 +47,11 @@ export class GraphSnapshot {
    *
    * @internal
    */
+<<<<<<< HEAD
   getNodeSnapshot(id: NodeId): Readonly<NodeSnapshot> | undefined {
+=======
+  getNodeSnapshot(id: NodeId): NodeSnapshot | undefined {
+>>>>>>> Address PR: rename getGraphNodeSnapshot
     return this._values[id];
   }
 
--- a/src/operations/SnapshotEditor.ts
+++ b/src/operations/SnapshotEditor.ts
@@ -128,13 +128,13 @@ export class SnapshotEditor {
    */
   private _mergePayloadValues(query: ParsedQuery, fullPayload: object): ReferenceEdit[] {
     const { entityIdForNode } = this._context;
-    const { parameterizedEdgeMap } = query.info;
+    const { dynamicEdgeMap } = query.info;
 
     const queue: MergeQueueItem[] = [{
       containerId: query.rootId,
       containerPayload: fullPayload,
       visitRoot: false,
-      edges: parameterizedEdgeMap,
+      edges: dynamicEdgeMap,
     }];
     const referenceEdits: ReferenceEdit[] = [];
     // We have to be careful to break cycles; it's ok for a caller to give us a
@@ -184,7 +184,6 @@ export class SnapshotEditor {
 
         if (isDynamicEdgeWithParameterizedArguments(edge)) {
           const edgeId = this._ensureParameterizedValueSnapshot(containerId, path, edge, query.variables!);
-
           // We walk the values of the parameterized edge like any other entity.
           //
           // EXCEPT: We re-visit the payload, in case it might _directly_
@@ -221,7 +220,8 @@ export class SnapshotEditor {
           // So, walk if we have new values, otherwise we're done for this
           // subgraph.
           if (nextNodeId) {
-            queue.push({ containerId: nextNodeId, containerPayload: payloadValue, visitRoot: false, edges: edge });
+            const updateEdge = edge instanceof DynamicEdge ? edge.children : edge;
+            queue.push({ containerId: nextNodeId, containerPayload: payloadValue, visitRoot: false, edges: updateEdge });
           }
           // Stop the walk for this subgraph.
           return true;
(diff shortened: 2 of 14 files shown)


src/util/references.ts
```diff
@@ -62,9 +63,16 @@ export function hasNodeReference(
...
+  const idx = getIndexOfGivenReference(references, id, path);
+  return idx !== -1;
+}
 
-  return false;
+/**
+ * Return index of { id, path } reference in references array.
+ * Otherwise, return -1.
+ */
+function getIndexOfGivenReference(references: NodeSnapshot.Reference[], id: NodeId, path?: PathPart[]): number {
```
yuit: may be we should use `Set` or `Map` for fast look up? my concern is that we will have to store both key and values of string. I guess we can do `Map<string, boolean> ` where key is an internalize string (`id_path[0]_path[1]`) and then re-parse it back

src/operations/SnapshotEditor.ts
```diff
@@ -152,10 +152,33 @@ export class SnapshotEditor {
...
         const nodeIsObject = isObject(nodeValue);
-        let nextNodeId = payloadIsObject ? entityIdForNode(payloadValue) : undefined;
+        let nextNodeId: string | undefined;
+        if (payloadIsObject) {
+          nextNodeId = entityIdForNode(payloadValue);
+          // If we can't find property "id", it doesn't mean that this
+          // payloadValue is not an entity trying to check if its "id"
+          // is alias. Note: we only need to walk one level!
+          // TODO: more generic approach to detect "id"
+          if (!nextNodeId) {
```
yuit: This may not needed (i.e disallow alias "id") - seem like `Apollo` cache also just use `dataIdFromObject` to get id (https://github.com/apollographql/apollo-client/blob/3b5045dfe74183fe060d2027e31317d890755b99/src/data/writeToStore.ts#L362) so if the user alias the "id" to be something else without updating this callback `Apollo` won't consider it a entity

src/GraphSnapshot.ts
```diff
@@ -27,7 +27,7 @@ export class GraphSnapshot {
    * Retrieves the value identified by `id`.
    */
   get(id: NodeId): any | undefined {
-    const snapshot = this.getSnapshot(id);
+    const snapshot = this.getGraphNodeSnapshot(id);
```
nevir: `getNodeSnapshot` IMO - it's a bit confusing otherwise

src/operations/SnapshotEditor.ts
```diff
@@ -152,10 +152,33 @@ export class SnapshotEditor {
...
+            const edgesChildren = dynamicEdges instanceof DynamicEdge ? dynamicEdges.children : dynamicEdges;
+            if (edgesChildren) {
+              for (const key in edgesChildren) {
+                const child = edgesChildren[key];
+                if (child instanceof DynamicEdge && child.fieldName === 'id') {
+                  nextNodeId = payloadValue[key];
+                  if (typeof nextNodeId === 'number' || typeof nextNodeId === 'string') {
+                    nextNodeId = `${nextNodeId}`;
```
nevir: Unfortunately, this isn't going to work, as the format of `id` is customizable by callers (often it's `${__typename}:${id}`, not just `${id}`).  In our case, we do that via `entityIdForNode` (which replaces `dataIdForNode`).  We also can't assume that all nodes with an `id` field should be identifiable (for example, our dashboards will want to blacklist certain types of nodes that _do_ have an `id` field).

I think we have two options here:

1. Translate the aliased keys _before_ calling `entityIdForNode`
2. Don't support alised keys when identifying nodes

I think I'm in favor of 2: simplifies our implementation, and it's a pretty ambiguous situation regardless

src/operations/SnapshotEditor.ts
```diff
@@ -249,11 +274,10 @@ export class SnapshotEditor {
           // values.
           if (nodeLength === payloadLength) return false;
 
-          const newArray = Array.isArray(nodeValue) ? nodeValue.slice(0, payloadLength) : [];
           // We will fill in the values as we walk, but we ensure that the
           // length is accurate, so that we properly handle empty values (e.g. a
           // value that contains only parameterized edges).
-          newArray.length = payloadLength;
+          const newArray = Array.isArray(nodeValue) ? nodeValue.slice(0, payloadLength) : new Array(payloadLength);
```
nevir: Doh! I totally forgot about that constructor 👍

src/operations/read.ts
```diff
@@ -112,39 +115,49 @@ export function _walkAndOverlayDynamicValues(
     }
 
     for (const key in edgeMap) {
-      let edge = edgeMap[key] as any;
+      if (key === "entityIdAlias") continue;
```
nevir: What if a payload has an `entityIdAlias` field in it?

src/util/tree.ts
```diff
@@ -67,6 +67,17 @@ export function walkPayload(
...
+  // node to interested property.
+  // i.e. {
+  //  user: {
+  //    name: "Bob",   -> path = ["user", "name"]
+  //    address: {     -> path = ["user", "address"]
+  //      city: "A",   -> path = ["user", "address", "city"]
+  //      state: "AB", -> path = ["user", "address", "state"]
+  //    }
+  //  }
+  // }
```
nevir: 👍  Might be good to illustrate an array in there too

test/unit/operations/write.ts
```diff
@@ -1594,4 +1594,532 @@ describe(`operations.write`, () => {
...
+
+          snapshot = write(config, empty, aliasQuery, {
+            user: {
+              id: 0,
+              FirstName: 'Foo',
+            },
+          }).snapshot;
+        });
+
+        it(`check write result`, () => {
```
nevir: nit: <code>it(\`only writes fields from the schema\`, () => {)</code>

nevir: (it's really helpful to give meaningful messages for tests, so that people can immediately narrow down problem areas from a list of failed tests)

src/operations/read.ts
```diff
@@ -112,39 +115,49 @@ export function _walkAndOverlayDynamicValues(
...
-        const childSnapshot = snapshot.getSnapshot(childId);
-        if (childSnapshot) {
-          edge = edge.children;
-          child = childSnapshot.node;
+      let fieldName = key;
+
+      if (edge instanceof DynamicEdge) {
+        // If exist filedName property then the current key is an alias
+        fieldName = edge.fieldName ? edge.fieldName : key;
+        nextEdge = edge.children;
```
nevir: I wonder if it'd actually be clearer if you just write over `edge` here, rather than juggle both `edge` and `nextEdge`?  we really don't want anyone referencing `edge` once this condition completes

src/util/ast.ts
```diff
@@ -92,57 +92,60 @@ export class DynamicEdge {
...
       // TODO: Memoize.
-      const fragmentEdges = buildParameterizedEdgeMap(fragments, fragment.selectionSet);
+      const fragmentEdges = buildDynamicEdgeMap(fragments, fragment.selectionSet);
       if (fragmentEdges) {
         edgeMap = { ...edgeMap, ...fragmentEdges };
       }
-    }
+    } else if (selection.kind === GraphqlNodeKind.Field) {
+      // if the current selection doesn't have any dynamic features
+      // but its children have dynamic features, we will hsot the DynamicEdgeMap
```
nevir: typo? "host"?

src/util/ast.ts
```diff
@@ -92,57 +92,60 @@ export class DynamicEdge {
...
+      // children. This is to save resources in walking
+      const currentKey: string = selection.alias ? selection.alias.value : selection.name.value;
+      let currentEdge: DynamicEdge | DynamicEdgeMap | undefined;
+      let parameterizedArguments: ParameterizedEdgeArguments | undefined;
+
+      if (selection.kind === GraphqlNodeKind.Field && selection.arguments && selection.arguments.length) {
+        parameterizedArguments = _buildParameterizedEdgeArgs(selection.arguments);
+      }
 
-    // TODO: inline fragments.
```
nevir: Let's keep this TODO around, we still don't support 'em yet

src/util/collection.ts
```diff
@@ -12,7 +12,7 @@ export function addToSet<T>(target: Set<T>, source: Iterable<T>): void {
 /**
  * An immutable deep set, where it only creates containers (objects/arrays) if
  * they differ from the _original_ object copied from - even if
- * `setValueForPath` is called against it multiple times.
+ * `_setValue` is called against it multiple times.
```
nevir: 👍 

VS Code should totally refactor my comments, yeesh!

src/util/references.ts
```diff
@@ -45,10 +43,13 @@ export function addNodeReference(
 ): void {
   let references = snapshot[direction];
   if (!references) {
-    references = (snapshot as any)[direction] = [];
+    references = snapshot[direction] = [];
   }
 
-  references.push({ id, path });
+  const idx = getIndexOfGivenReference(references, id, path);
+  if (idx === -1) {
```
nevir: I think I'd rather we throw an error in this case; AFAIK all places that call `addNodeReference` already have to guard against this case, or at least thought they were

src/util/references.ts
nevir: I suspect that iterating will be faster overall; these lists aren't very long in general - let's do a perf pass once we've got functionality in

src/operations/read.ts
yuit: This is me trying to do something removed :)

src/operations/read.ts
yuit: Agree about trying to keep track of both `edge` and `nextEdge` can be bug prone

src/util/ast.ts
yuit: :D

src/util/references.ts
yuit: This may not be true as now we have alias so it is possible that we have already add an alias (that gets de-alias) then encounter same field with different alias or in its true form.  I do think we shouldn't call the check before call this function (so we don't have to do unnecessary loop over array)

src/util/tree.ts
yuit: not sure what you mean by illustrate an array?

PR #88, opened 2017-08-28
Extract DynamicEdge & friends out to a separate module

nevir:


yuit:
Alt Text FINALLY find a way to gif :) this will be easten by ur cat gif

PR #89, opened 2017-08-28
Don't allow undefined keys in dynamic edge maps

nevir:


yuit:
Oh that is right. we should never have undefined (caz we shouldn't be creating anything if it doesn't need

PR #91, opened 2017-08-28
Rename edge to field

nevir:
Field is the term used throughout the GraphQL spec and within other projects.  Let's get our terminology in line, to reduce confusion


yuit:
Alt Text

PR #93, opened 2017-08-28
Extract several bits of metadata about the variables used by queries

nevir:
This is in preparation for:

* Supporting optional parameters/values (e.g. substituting their values w/ defaults)
* Expanding variables at query parse time, rather than tree walking time


src/DynamicField.ts
```diff
@@ -176,3 +169,12 @@ export function expandFieldArguments(args: FieldArguments, variables: object = {
 
   return expanded;
 }
+
+/**
+ *
+ */
+export function variablesInDynamicFieldMap(fieldMap: DynamicFieldMap): Set<string> {
```
yuit: Q: is this just place holder?

src/util/ast.ts
```diff
@@ -23,6 +29,68 @@ export function getOperationOrDie(document: DocumentNode): OperationDefinitionNo
   return operations[0];
 }
 
+/**
+ * Returns the names of all variables declared by the operation.
+ */
+export function variablesInOperation(operation: OperationDefinitionNode): Set<string> {
```
yuit: nit: I think it may be helpful to prefix "getter" type function with "get". For example `getVariablesInOperation`

src/util/ast.ts
```diff
@@ -23,6 +29,68 @@ export function getOperationOrDie(document: DocumentNode): OperationDefinitionNo
...
+
+      const { defaultValue } = definition;
+      defaults[definition.variable.name.value] = isObject(defaultValue) ? valueFromNode(defaultValue) : null;
+    }
+  }
+
+  return defaults;
+}
+
+export function _defaultValueFromVariable(node: VariableNode) {
```
yuit: Q: why did we export this function ?

src/DynamicField.ts
```diff
@@ -59,7 +58,18 @@ export class VariableArgument {
...
  */
-export function buildDynamicFieldMap(fragments: FragmentMap, selectionSet?: SelectionSetNode): DynamicFieldMap | undefined {
+export function buildDynamicFieldMap(fragments: FragmentMap, selectionSet?: SelectionSetNode) {
+  const variables = new Set<string>();
+  const fieldMap = _buildDynamicFieldMap(variables, fragments, selectionSet);
+
+  return { variables, fieldMap };
+}
+
+export function _buildDynamicFieldMap(
```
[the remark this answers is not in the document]
yuit: nit: I think `buildDynamicFieldMap` and `_buildDynamicFieldMap ` is a bit confusing here because `DynamicFieldMap` is now both just `fieldMap` and `{ variables, fieldMap }`

src/DynamicField.ts
```diff
@@ -124,28 +134,11 @@ function _buildFieldArgs(argumentsNode: ArgumentNode[]): FieldArguments {
...
-    for (const field of node.fields) {
-      value[field.name.value] = _valueFromNode(field.value);
-    }
-    return value;
-  }
-  default:
-    return node.value;
-  }
+function _valueFromNode(variables: Set<string>, node: ValueNode): any {
+  return valueFromNode(node, ({ name: { value } }) => {
```
yuit: note: no need to do it now. but just wonder if we should lift this callback out (as we keep creating new function object everytime -> can be expensive for garbage)

src/context/QueryInfo.ts
```diff
@@ -18,21 +26,67 @@ export class QueryInfo {
...
+
+  private _assertAllVariablesDeclared(messages: string[], declaredVariables: Set<string>) {
+    for (const name of this.variables) {
+      if (!declaredVariables.has(name)) {
+        messages.push(`Variable $${name} is used, but not declared`);
+      }
+    }
+  }
+
+  private _assertAllVariablesUsed(messages: string[], declaredVariables: Set<string>) {
```
yuit: For Yui future ref: https://facebook.github.io/graphql/#sec-All-Variables-Used

nevir: `graphql-tag` unfortunately doesn't enforce it, hence our extra check :(

src/DynamicField.ts
nevir: Doh, yeah, didn't mean to include that 👍

src/DynamicField.ts
nevir: I'm not sure if we can very easily (since we want `variables` in the closure) :(

src/util/ast.ts
nevir: habit - and useful for unit testing "private" module methods; but leaving private until we need to!

src/util/ast.ts
nevir: I mean, it's pretty much true of all of the pure functions we have; hmm

src/context/QueryInfo.ts
jbaxleyiii: @nevir we could fix this 👍

nevir: Hah! 👍

PR #96, opened 2017-08-30
Assert JSON-serializable types where we can

nevir:
We should be handling JSON-serializable types…

* …when given a payload from the server
* …from Apollo when writing values
* …when storing values
* …when returning values

### DIFF
--- a/src/apollo/ApolloQueryable.ts
+++ b/src/apollo/ApolloQueryable.ts
@@ -1,3 +1,4 @@
+import { JsonObject } from '../primitive';
 import { Queryable } from '../Queryable';
 
 import * as interfaces from './interfaces';
@@ -42,23 +43,23 @@ export abstract class ApolloQueryable {
 
   readFragment<FragmentType>(options: interfaces.Cache.ReadFragmentOptions, optimistic?: true): FragmentType | null {
     // TODO: Support nested fragments.
-    const query = toQuery(options.fragment, options.variables);
+    const query = toQuery(options.fragment, options.variables as JsonObject);
     return this._queryable.read(query, optimistic).result;
   }
 
   writeResult(options: interfaces.Cache.WriteResultOptions): void {
-    const query = toQuery(options.document, options.variables, options.dataId);
+    const query = toQuery(options.document, options.variables as JsonObject, options.dataId);
     this._queryable.write(query, options.result);
   }
 
   writeQuery(options: interfaces.Cache.WriteQueryOptions): void {
-    const query = toQuery(options.query, options.variables);
+    const query = toQuery(options.query, options.variables as JsonObject);
     this._queryable.write(query, options.data);
   }
 
   writeFragment(options: interfaces.Cache.WriteFragmentOptions): void {
     // TODO: Support nested fragments.
-    const query = toQuery(options.fragment, options.variables, options.id);
+    const query = toQuery(options.fragment, options.variables as JsonObject, options.id);
     this._queryable.write(query, options.data);
   }
 }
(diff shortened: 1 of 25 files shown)


test/unit/context/CacheContext/entityTransformation.ts
```diff
@@ -279,10 +280,10 @@ describe(`context.CacheContext`, () => {
 
       it(`get information through helper methods`, () => {
         const { result } = read(entityTransformerContext, viewerQuery, snapshot);
-        expect(result.user.getName()).to.eq('Bob');
-        expect(result.user.getId()).to.eq('0');
-        expect(result.user.getJustPhoneNumber()).to.eq(1234);
-        expect(result.user.getCity()).to.eq('AA');
+        expect((result as any).user.getName()).to.eq('Bob');
```
nevir: I'm debating having `read` return `any` by default to make this a bit more ergonomic

src/nodes/ParameterizedValueSnapshot.ts
```diff
@@ -1,5 +1,10 @@
+import { JsonValue, NestedArray, NestedObject } from '../primitive';
```
yuit: nit: since this is jus timport to be reexport (we can use re-export syntax `export { JsonValue, NestedArray, NestedObject } from '../primitive';` instead and remove export below

src/apollo/ApolloQueryable.ts
```diff
@@ -42,23 +43,23 @@ export abstract class ApolloQueryable {
 
   readFragment<FragmentType>(options: interfaces.Cache.ReadFragmentOptions, optimistic?: true): FragmentType | null {
     // TODO: Support nested fragments.
-    const query = toQuery(options.fragment, options.variables);
-    return this._queryable.read(query, optimistic).result;
+    const query = toQuery(options.fragment, options.variables as JsonObject);
+    return this._queryable.read(query, optimistic).result as any;
```
yuit: Q: why we cast this to any (but return type is `FragmentType`)?

nevir: TS thinks they're incompatible:

```ts
src/apollo/ApolloQueryable.ts(47,12): error TS2352: Type 'string | number | boolean | NestedArray<JsonScalar> | NestedObject<JsonScalar> | null | undefined' cannot be converted to type 'FragmentType'.
  Type 'NestedObject<JsonScalar>' is not comparable to type 'FragmentType'.
```

and I don't want to change the interface, since it'll eventually move out of our package

src/nodes/ParameterizedValueSnapshot.ts
nevir: No dice - it has to be `import`ed as well, as it's a transitive dep of the current module

src/operations/SnapshotEditor.ts
```diff
@@ -159,8 +159,8 @@ export class SnapshotEditor {
       walkPayload(containerPayload, container, fields, visitRoot, (path, payloadValue, nodeValue, dynamicFields) => {
         const payloadIsObject = isObject(payloadValue);
         const nodeIsObject = isObject(nodeValue);
-        let nextNodeId = payloadIsObject ? entityIdForNode(payloadValue) : undefined;
-        const prevNodeId = nodeIsObject ? entityIdForNode(nodeValue) : undefined;
+        let nextNodeId = payloadIsObject ? entityIdForNode(payloadValue as JsonObject) : undefined;
```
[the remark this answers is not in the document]
nevir: TS isn't inferring that they're an object here (it still thinks they're a `JsonValue`, even though `payloadIsObject` implies `JsonObject`)

PR #97, opened 2017-08-31
Expand variables at query parse time, and support for optional/default arguments

### LINKED ISSUE #85: Flatten variables when compiling queries
Currently, we interpolate variables at each field that requests them - instead, we could do this once when building the dynamic edge map

nevir:
Fixes #85 

We now have two "kinds" of `DynamicField`s - those with `VariableArgument`s in their arguments, and those without.  The result of parsing the AST yields (potentially) fields with variables.  

However, when parsing a query (`CacheContext#parseQuery`), we immediately substitute in the values for those variables, to minimize work when reading/writing.  This also allows us to trivially support optional & default field arguments

### DIFF
--- a/src/DynamicField.ts
+++ b/src/DynamicField.ts
@@ -7,29 +7,29 @@ import { // eslint-disable-line import/no-extraneous-dependencies, import/no-unr
 import { JsonObject, JsonScalar, JsonValue, NestedObject, NestedValue } from './primitive';
 import { FragmentMap, isObject, valueFromNode } from './util';
 
+export type JsonAndArgs = JsonScalar | VariableArgument;
+
 /**
  * Represent dynamic information: alias, parameterized arguments, directives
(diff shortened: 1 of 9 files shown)


src/DynamicField.ts
```diff
@@ -4,40 +4,38 @@ import { // eslint-disable-line import/no-extraneous-dependencies, import/no-unr
...
+import { FragmentMap, isObject, valueFromNode } from './util';
+
+export type JsonAndArgs = JsonScalar | VariableArgument;
 
 /**
  * Represent dynamic information: alias, parameterized arguments, directives
  * (if existed) of NodeSnapshot in GraphSnapshot.
  */
-export class DynamicField {
+export class DynamicField<TArgTypes = JsonScalar, TChildArgTypes = JsonScalar> {
```
nevir: Split generic so that typescript can properly infer when there are no explicit template types provided

PR #98, opened 2017-08-31
Handle duplicated reference updates (arrays, nested, etc)

nevir:
When we receive a payload that contains duplicated reference edits (e.g. duplicate subgraphs), ensure we handle it properly.

### DIFF
--- a/src/util/references.ts
+++ b/src/util/references.ts
@@ -18,9 +18,7 @@ export function removeNodeReference(
   path?: PathPart[],
 ): boolean {
   const references = snapshot[direction];
-  if (!references) {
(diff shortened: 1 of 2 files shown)


src/util/references.ts
```diff
@@ -18,9 +18,7 @@ export function removeNodeReference(
   path?: PathPart[],
 ): boolean {
   const references = snapshot[direction];
-  if (!references) {
-    throw new Error(`Inconsistent GraphSnapshot: Expected snapshot to have ${direction} references`);
-  }
```
nevir: Turns out it's expected, as we'll end up with multiple `referenceEdit` entries in `SnapshotEditor`

PR #110, opened 2017-09-01
Track all queries we've previously written, and ensure that they come back as complete

nevir:
We've been seeing a few cases where we've written a payload to the cache, but then had it marked incomplete when immediately read.

This should help to track those cases down (and also recover from them)

### DIFF
--- a/src/Cache.ts
+++ b/src/Cache.ts
@@ -27,6 +27,10 @@ export class Cache implements Queryable {
   /** All active query observers. */
   private _observers: QueryObserver[] = [];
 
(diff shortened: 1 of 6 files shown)


src/Cache.ts
```diff
@@ -77,10 +77,12 @@ export class Cache implements Queryable {
    *
    * If a changeId is provided, the transaction will be recorded as an
    * optimistic update.
+   *
+   * Returns whether the transaction was successful.
```
nevir: Sneaking this in while I'm here

PR #112, opened 2017-09-02
Garbage collect parameterized edges in nodes that are removed from arrays

nevir:
* Makes `path` a required property on all references
* We now inspect the kind of node being traversed when rebuilding references (only rebuild entity references!)
* Fixes garbage collection of parameterized edges contained within arrays

### DIFF
--- a/src/operations/SnapshotEditor.ts
+++ b/src/operations/SnapshotEditor.ts
@@ -96,14 +96,15 @@ export class SnapshotEditor {
     // all references that have changed.  Reference changes are applied later,
     // once all new nodes have been built (and we can guarantee that we're
     // referencing the correct version).
-    const referenceEdits = this._mergePayloadValues(parsed, payload);
+    const { referenceEdits, orphanedNodeIds } = this._mergePayloadValues(parsed, payload);
 
     // Now that we have new versions of every edited node, we can point all the
     // edited references to the correct nodes.
     //
     // In addition, this performs bookkeeping the inboundReferences of affected
     // nodes, and collects all newly orphaned nodes.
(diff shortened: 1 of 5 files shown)


yuit:
just found this. Looking right now

src/operations/SnapshotEditor.ts
```diff
@@ -331,17 +335,17 @@ export class SnapshotEditor {
...
     while (queue.length) {
       const nodeId = queue.pop()!;
       const snapshot = this.getNodeSnapshot(nodeId);
+      if (!(snapshot instanceof EntitySnapshot)) continue;
       if (!snapshot || !snapshot.inbound) continue;
 
       for (const { id, path } of snapshot.inbound) {
-        if (!path) continue;
         this._setValue(id, path, snapshot.node, false);
         if (this._rebuiltNodeIds.has(id)) continue;
```
yuit: Q: should this be called before call `this._setValue`?

yuit: Q: should we call `this._rebuiltNodeIds.add(nodeId`)` after the for loop?

yuit:
Alt Text

src/operations/SnapshotEditor.ts
nevir: > Q: should this be called before call this._setValue?

Newp - they're two independent pieces of code, really:

1. Write the current node to the inbound node
2. Enqueue the inbound node, if we haven't previously

> Q: should we call this._rebuiltNodeIds.add(nodeId)` after the for loop?

Nay, it's guaranteed to be part of the set if we're iterating it (we always add to `_rebuiltNodeIds` prior to pushing to the queue)

PR #113, opened 2017-09-03
update typescript and fix error

jbaxleyiii:
Since TypeScript 2.5 came out, the `^2.4.2` installs it by default. With that version, there was a TypeError for calling `Object.keys` on a possible non object. This force casts the type def for it.

### DIFF
--- a/src/operations/SnapshotEditor.ts
+++ b/src/operations/SnapshotEditor.ts
@@ -2,7 +2,7 @@ import { CacheContext } from '../context';
 import { DynamicField, DynamicFieldWithArgs, DynamicFieldMap, isDynamicFieldWithArgs } from '../DynamicField';
 import { GraphSnapshot } from '../GraphSnapshot';
 import { EntitySnapshot, NodeSnapshot, ParameterizedValueSnapshot, cloneNodeSnapshot } from '../nodes';
-import { JsonObject, PathPart } from '../primitive';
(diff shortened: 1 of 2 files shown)


src/operations/SnapshotEditor.ts
```diff
@@ -278,7 +278,11 @@ export class SnapshotEditor {
         // TODO: Rather than detecting empty objects directly (which should
         // never occur for GraphQL results, and only for custom types), we
         // should be walking the selection set of the query.
-        } else if (payloadIsObject && !Object.keys(payloadValue).length && (!nodeIsObject || Object.keys(payloadValue).length)) {
+        } else if (
+          payloadIsObject &&
+          !Object.keys((payloadValue as NestedObject<JsonScalar>)).length &&
```
nevir: Total nit for this case (I'll make the change) - you can use [`JsonObject`](https://github.com/convoyinc/apollo-cache-hermes/blob/master/src/primitive.ts#L51) for this guy

src/operations/SnapshotEditor.ts
```diff
@@ -278,7 +278,11 @@ export class SnapshotEditor {
...
+          (!nodeIsObject || Object.keys((payloadValue as NestedObject<JsonScalar>)).length)
+        ) {
```
nevir: Man, can't wait to get rid of this hack :P

PR #122, opened 2017-09-05
Copy Prototype when read

yuit:
This is a fix to get nacho-dashboard to use entityTransformer.

### DIFF
--- a/src/operations/read.ts
+++ b/src/operations/read.ts
@@ -173,8 +173,26 @@ export function _walkAndOverlayDynamicValues(
 function _wrapValue(value: JsonValue): any {
   if (value === undefined) return {};
   if (Array.isArray(value)) return [...value];
-  if (isObject(value)) return { ...value };
+  if (isObject(value)) {
+    const newValue = { ...value };
+    _setPrototypeOf(newValue, _getPrototypeOf(value));
+    newValue.__proto__ = value.__proto__;
+    return newValue;
+  }
   return value;
+}
+
+function _getPrototypeOf(obj: { __proto__?: object } ): object {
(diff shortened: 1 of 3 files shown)


src/operations/read.ts
```diff
@@ -173,10 +173,28 @@ export function _walkAndOverlayDynamicValues(
...
+    const newValue = { ...value };
+    _setPrototypeOf(newValue, _getPrototypeOf(value));
+    newValue.__proto__ = value.__proto__;
+    return newValue;
+  }
   return value;
 }
 
+function _getPrototypeOf(obj: { __proto__?: object } ): object {
+  // Object.getPrototypeOf is only available in ES5 and beyond
```
nevir: We (and Apollo) are assuming a polyfilled environment - use the modern way!

src/operations/read.ts
```diff
@@ -173,10 +173,28 @@ export function _walkAndOverlayDynamicValues(
 function _wrapValue(value: JsonValue): any {
   if (value === undefined) return {};
   if (Array.isArray(value)) return [...value];
-  if (isObject(value)) return { ...value };
+  if (isObject(value)) {
+    const newValue = { ...value };
+    _setPrototypeOf(newValue, _getPrototypeOf(value));
+    newValue.__proto__ = value.__proto__;
+    return newValue;
+  }
```
nevir: IMO: Rather than copying just the prototype, you actually want to call the entity transformer here (if this is an entity).

For example, if we end up moving object freezing (in dev mode) into the transformer, it would also need to be able to freeze these

nevir:
Sorry, one last thing - let's add a test or two for this case!

yuit:
@nevir Good catch 💃

PR #123, opened 2017-09-05
Refactor SnapshotEditor a bit

nevir:
This is a slight perf optimization for optimistic updates - we move the inbound reference rewriting to commit(), so that it only occurs once when we merge multiple payloads in the same transaction.

### DIFF
--- a/src/operations/SnapshotEditor.ts
+++ b/src/operations/SnapshotEditor.ts
@@ -314,8 +314,20 @@ export class SnapshotEditor {
    * Commits the transaction, returning a new immutable snapshot.
    */
   commit(): EditedSnapshot {
+    return {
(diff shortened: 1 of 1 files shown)


src/operations/SnapshotEditor.ts
```diff
@@ -375,48 +411,21 @@ export class SnapshotEditor {
...
-        snapshots[id] = newSnapshot;
-      }
-    }
-
-    return {
-      snapshot: new GraphSnapshot(snapshots),
-      editedNodeIds: this._editedNodeIds,
-      writtenQueries: this._writtenQueries,
-    };
+  private _getNodeSnapshot(id: NodeId) {
```
nevir: Me being nitpicky

PR #129, opened 2017-09-11
Restore nearest deps from cache, if there are none for that exact one

nevir:
Leverage Circle's prefix lookup to find the most recently installed deps if we don't have a cache for the current configuration of `package.json`


nevir:
Nope, don't like this.  Behavior is too ambiguous (files stick around indefinitely)

PR #130, opened 2017-09-12
update to cache-core

jbaxleyiii:
@nevir @yuit this updates to the latest stable API for the Apollo Cache as well as factors out some copied utilities in favor of the new apollo-utilities package on npm.

One thing to note is the name change for the main ApolloCache interface. Since the cache-core is ApolloCache, I thought it was best to name this implementation Hermes instead of overwriting the class namespace. Thoughts?

I'm still figuring my way around so please let me know thoughts!


jbaxleyiii:
@nevir I'm going to update the package name here (apollo-cache-core => apollo-cache) 👍

src/apollo/Queryable.ts
```diff
@@ -1,28 +1,34 @@
+/* eslint-disable class-methods-use-this */
```
nevir: You heathen!  break 'em out into helper funcs, or just disable it for the specific method

src/apollo/Queryable.ts
```diff
@@ -1,28 +1,34 @@
...
 /**
  * Apollo-specific interface to the cache.
  */
-export abstract class ApolloQueryable {
+export abstract class ApolloQueryable implements DataProxy {
   /** The underlying Hermes cache. */
   protected abstract _queryable: Queryable;
 
-  diffQuery(options: interfaces.Cache.DiffQueryOptions): interfaces.Cache.DiffResult {
+  diff<T = any>(options: Cache.DiffOptions): Cache.DiffResult<T | any> {
```
nevir: Hmm, why `T | any` for the result? (particularly if the template parameter already defaults to `any`)

src/apollo/Queryable.ts
```diff
@@ -1,28 +1,34 @@
...
     const query = toQuery(options.query, options.variables);
     const { result, complete } = this._queryable.read(query, options.optimistic);
     if (options.returnPartialData === false && !complete) {
       // TODO: Include more detail with this error.
       throw new Error(`diffQuery not satisfied by the cache.`);
     }
 
-    return { result, isMissing: !complete };
+    return { result, complete };
```
nevir: woot!

src/apollo/index.ts
```diff
@@ -1 +1,4 @@
-export { ApolloCache } from './ApolloCache';
+import { Hermes } from './Hermes';
+export { Hermes };
+export default Hermes;
```
nevir: nit: let's drop the export `default` (particularly for private modules like this) - rationale: `default` exports construct a new symbol to reference that value, making it difficult to refactor the name of something

src/util/ast.ts
```diff
@@ -1,47 +1,21 @@
...
-
-  return names;
-}
+export {
+  getOperationDefinitionOrDie as getOperationOrDie,
+  variablesInOperation,
+  valueFromNode,
+  FragmentMap,
+  addTypenameToDocument,
+} from 'apollo-utilities';
```
nevir: glorious!

src/Cache.ts
```diff
@@ -57,7 +57,7 @@ export class Cache implements Queryable {
    * Registers a callback that should be triggered any time the nodes selected
    * by a particular query have changed.
    */
-  watch(query: Query, callback: () => void): () => void {
+  watch(query: Query, callback: (newData: any) => void): () => void {
```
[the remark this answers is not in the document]
jbaxleyiii: agreed! I *think* I added that to the latest release. If not I'll do it!

scripts/test-unit-coverage.sh
```diff
@@ -13,7 +13,7 @@ fi
 run test:unit -- --coverage "${OPTIONS[@]}"
 
 if [[ "${CI}" == "" ]]; then
-  open ./output/test:unit/index.html
+  open ./output/test-unit/index.html
```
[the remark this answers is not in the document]
jbaxleyiii: sure thing!

src/apollo/Hermes.ts
```diff
@@ -0,0 +1,57 @@
...
+  /** The underlying Hermes cache. */
+  protected _queryable: Cache;
+
+  constructor(configuration?: CacheContext, intitialState?: GraphSnapshot) {
+    super();
+    this._queryable = new Cache(configuration, intitialState);
+  }
+
+  getData(): any {
+    return this._queryable.getSnapshot().baseline;
```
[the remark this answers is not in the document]
jbaxleyiii: `getOptimisticData ` is currenty used by the dev tools. `getData` was used for getting the store for SSR. Honestly I'll going to try to kill both of them and maybe make an optional API for SSR? `.extract()`?

However, for SSR, having a way to serialize and boot up from will be critical for a number of users. Thoughts on how I should go about implementing it?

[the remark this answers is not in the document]
nevir: Ah!  I think for that, we can have a unified API for cache implementers - I think something like:

```
interface ApolloCache<TSerialized> {
  /**
   * Replaces existing state in the cache (if any) with the values expressed by
   * `serializedState`.
   *
   * Called when hydrating a cache (server side rendering, or offline storage),
   * and also (potentially) during hot reloads.
   */
  restore(serializedState: TSerialized): void;

  /**
   * Exposes the cache's complete state, in a serializable format for later restoration.
   */
  save(optimistic?: boolean): TSerialized;
}
```

PR #135, opened 2017-09-20
Write the cache by using query selectionSet as a guide

yuit:
- [ ] Add fragment query test

- [ ] Add read & write test for array with holes

- [ ] Add test for array which has an element that miss property

- [ ] turn on unusedParameter in tsconfig


nevir:
TODO: Integrate changes from #137

PR #158, opened 2017-10-11
1 - Split write/cyclicReferences tests

yuit:
Part of #136 

Note: **REMOVE** following cases because JSON can't have cycle reference

- "cyclic values in payloads"
- "cyclic references in payloads"


test/unit/operations/write/cyclicReferences/editGraph.ts
```diff
@@ -0,0 +1,87 @@
...
+// It just isn't very fruitful to unit test the individual steps of the write
+// workflow in isolation, given the contextual state that must be passed around.
+describe(`operations.write`, () => {
+
+  const context = new CacheContext(strictConfig);
+  const empty = new GraphSnapshot();
+
+  describe(`cyclic references`, () => {
+
+    describe(`when editing`, () => {
```
nevir: Let's merge these `describe`s into one line, to fit that rule.  Might also want to do it with the top level `describe` too?

yuit: Talk off line, will merge all these PR and then clean up with another round of PR

PR #166, opened 2017-10-12
Move addTypename into transformDocument

nevir:
Apollo calls this prior to hitting the cache up for any operation.  This allows us to move to a world where injecting typenames doesn't mutate the original document.


yuit:
much niceee!

PR #167, opened 2017-10-12
Write tests cleanup

yuit:
Part of #136. This PR is best **reviewed by commit** as each commit is equal to an update of the folder. 

What happen in this PR are:

- simplifying `describe` clause. We only have two layers of describe clauses - one for specifying the tested operation such as write, read etc. the other is a test scenario itself.

- rename file to have better name

Some guidelines I try to stick with:

- `describe` description is to describe scenario, majority are just type of payload or query.

- without verb to describe an act, we assume tests are just doing write operation.

- If tests do more than just write, use verb to specify such action in the `describe` description such as "updates" etc.

- anything that is an entity in the system is referred as `references` as `entity` will be represent as reference off a root query

- anything that is non-entity is referred as `values`

- All setup should be in `beforeAll` clause; `it` clause should just do the test


test/unit/operations/write/basic/invalidNestedValues.ts
```diff
@@ -20,7 +20,7 @@ describe(`operations.write`, () => {
     }
   }`);
 
-  describe(`invalid write on a root, expect an object`, () => {
+  describe(`invalid nested values, expect an object`, () => {
```
nevir: "expect an object" is still a verb phrase - maybe "scalar in place of object value"?

test/unit/operations/write/basic/updatesInnerNodes.ts
```diff
@@ -23,7 +23,7 @@ describe(`operations.write`, () => {
     }
   }`);
 
-  describe(`write to inner nodes`, () => {
+  describe(`updates inner nodes`, () => {
```
nevir: "inner node updates"? maybe

Issue #157, opened 2017-10-11
[Infra] Consolidate common functions among test files 


yuit:
There are a lot of duplicate code across test files. We should consolidate them and write helper functions/classes etc. to do all test setup work.

PR #169, opened 2017-10-12
Create helpers for write operation test

yuit:
Part of #157, - create helper functions for common operation: creating EditedSnapshot and updating one. 

Only `basic` & `cyclicReferences` folder are updated as a study. I will do the rest after finishing the serialization work.


test/helpers/write.ts
```diff
@@ -0,0 +1,37 @@
...
+import { strictConfig } from './context';
+import { query } from './graphql';
+
+export type QueryObject = {
+  gqlString: string,
+  gqlVariables?: JsonObject,
+  rootId?: NodeId,
+};
+
+export function createBaselineEditedSnapshot(
```
nevir: Nit: `createSnapshot` is probably sufficient here, or `createBaselineSnapshot` (`EditedSnapshot` is not a great name for that return type)

test/helpers/write.ts
```diff
@@ -0,0 +1,37 @@
...
+  { gqlString, gqlVariables, rootId }: QueryObject,
+  payload: JsonObject,
+  cacheConfig: CacheContext.Configuration = strictConfig
+): EditedSnapshot {
```
nevir: Ergonomically, I think it's going to be a lot simpler if we can remove some of these arguments:

 * `rootId` will probably never be overridden (and maybe should be exposed via an "advanced" form of the method if we do need to)
* variables could be inlined into the query

IMO:

```ts
export function createSnapshot(
  gqlString: string, 
  payload: JsonObject, 
  cacheConfig: CacheContext.Configuration = strictConfig,
): EditedSnapshot {

yuit: I like that idea. Originally I had similar function signature as above but at small number of places we specify `rootId` -> hence it probably makes more send for "advanced" function

yuit:
Will merge for now.  but feel free to let me know if the signature is too much :)

PR #176, opened 2017-10-18
Resolver redirects

### LINKED ISSUE #47: Cache redirects
Need to support cache redirects.

Motivation is to support queries like:

```graphql
query getAUser($id: ID!) {
  user(id: $id) { ... }
}
```

and return the entity with that id, no matter where it was retrieved from

nevir:
Resolves #47 (also see [design doc](https://docs.google.com/document/d/1x2tQWTJbchfoU8i56NTxwsnqnpDJMQvPVKQ5JiHbSco/))

Note that this only works for parameterized fields, and only allows you to point to entities

### DIFF
--- a/src/context/CacheContext.ts
+++ b/src/context/CacheContext.ts
@@ -79,8 +79,11 @@ export namespace CacheContext {
     freeze?: boolean;
 
     /**
-     * Parameterized fields that should redirect to other locations in the cache
-     * when there is no value currently cached for their location.
+     * Parameterized fields that should redirect to entities in the cache when
+     * there is no value currently cached for their location.
+     *
+     * Note that you may only redirect to _entities_ within the graph.
+     * Redirection to arbitrary nodes is not supported.
      */
     resolverRedirects?: ResolverRedirects;
   }
--- a/src/context/CacheContext.ts
+++ b/src/context/CacheContext.ts
@@ -25,6 +25,17 @@ export namespace CacheContext {
   }
 
   /**
+   * Expected to return an EntityId or undefined, but we loosen the restrictions
+   * for ease of declaration.
+   */
+  export type ResolverRedirect = (args: JsonObject) => any;
+  export type ResolverRedirects = {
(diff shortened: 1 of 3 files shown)


test/unit/operations/read/redirects.ts
```diff
@@ -0,0 +1,76 @@
...
+      });
+
+      const { result, complete } = read(context, { ...rootRedirectQuery, variables: { id: 1 } }, snapshot);
+      expect((result as any).thing).to.deep.eq({ __typename: 'Thing', id: 111, name: 'Other One' });
+      expect(complete).to.eq(true);
+    });
+
+    it(`supports redirects to nowhere`, () => {
+      const { result, complete } = read(context, { ...rootRedirectQuery, variables: { id: 123 } }, baseSnapshot);
+      expect((result as any).thing).to.deep.eq(undefined);
```
yuit: Should this be `null` instead?

src/context/CacheContext.ts
```diff
@@ -25,6 +25,17 @@ export namespace CacheContext {
   }
 
   /**
+   * Expected to return an EntityId or undefined, but we loosen the restrictions
```
yuit: We should document here that that this redirect only from parameterized field to Entity only

src/context/CacheContext.ts
```diff
@@ -66,6 +77,12 @@ export namespace CacheContext {
      * Defaults to true unless process.env.NODE_ENV === 'production'
      */
     freeze?: boolean;
+
+    /**
+     * Parameterized fields that should redirect to other locations in the cache
```
yuit: I think we should make sure that "other locations" is strictly an entity

test/unit/operations/read/redirects.ts
nevir: Nope!  `undefined` indicates the value hasn't been populated in the cache - `null` would indicate that we _have_ retrieved it, but it's just not present (on the server)

PR #182, opened 2017-10-19
Fix field aliases when computing read completeness

jamesreggio:
`walkOperation` and the general computation of read completeness were broken when it concerned complex fields with alias names. This patches a couple of gaps.

Tests included, non-breaking.

### DIFF
--- a/src/operations/read.ts
+++ b/src/operations/read.ts
@@ -219,7 +219,8 @@ export function _visitSelection(
     }
 
     for (const field of fields) {
-      if (!(field.name.value in value)) {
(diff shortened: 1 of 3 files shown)


src/operations/read.ts
```diff
@@ -219,7 +219,8 @@ export function _visitSelection(
     }
 
     for (const field of fields) {
-      if (!(field.name.value in value)) {
+      const nameNode = field.alias || field.name;
+      if (!(nameNode.value in value)) {
```
nevir: Hell.  Good catch!

Issue #153, opened 2017-10-10
Support for union queries


nevir:
We don't support `... on Type {` queries currently.  But it shouldn't be too difficult to support, I think. 

Rough outline of the work:

1. When [parsing queries](https://github.com/convoyinc/apollo-cache-hermes/blob/master/src/ParsedQueryNode.ts), we need to walk `InlineFragment` AST nodes
  * Most likely, they are added as another field node (but explicitly flagged as being a union field)
2. When [writing fields](https://github.com/convoyinc/apollo-cache-hermes/blob/master/src/operations/SnapshotEditor.ts), any union field present in the payload should be written normally
3. When [reading queries](https://github.com/convoyinc/apollo-cache-hermes/blob/master/src/operations/read.ts) there should be no work required for unions with only static fields

Potential challenges:

* Parameterized fields contained by a union query may need special handling during  (and `__typename` or schema introspection)

anthonychung14:
I'd be interested in helping with this, though I'm just now getting onboarded with the source code. Our team would benefit greatly from incorporating the improvements you've made with the apollo cache.

Here's how I understand the lib + #1 to work
- Currently, reading an inline fragment would result in a "failed lookup" since it doesn't know that the fragment needs to be rebuilt (later on?). I'm assuming this is why you mentioned flagging them as union fields.

- It seems like here detects and flattens fragments into the nodeMap. This implies to me that fragments are already (in a sense) being "walked" and they only need to be flagged as unions.
https://github.com/convoyinc/apollo-cache-hermes/blob/master/src/ParsedQueryNode.ts#L148

- So, would we then want to adjust how `_mergeNodes` works to complete this flagging?

jamesreggio:
Hi @anthonychung14 — welcome to the project.

I've implemented 95% of the support needed for inline fragments over in #183. If you'd like to begin using it today, I'd suggest you incorporate my branch into a private fork.

(The comment thread on that PR outlines the remaining work needed to get support to 100%, and it's not trivial — it involves having the entire schema available on the client to understand which fragment was matched. This remaining work is mostly in the realm of validation, so it doesn't need to be finished to begin using the branch today.)

jamesreggio:
We've merged support for inline fragments in #183. In the process of implementing it, it became clear that the 'hard' part of this is handling multiple fragments for different types on a given object. This is a common scenario when querying a union type (because you'll likely have one fragment for potential type in the union), but it also crops up when querying object interfaces.

[...]

PR #183, opened 2017-10-19
Add support for inline fragments

jamesreggio:
Related to #153.

These simple changes add basic support for inline fragments (also known as union queries).

These changes were enough to make our app work, which makes heavy use of inline fragments and union types. Obviously, tests need to be written, but I'm not certain I'll have time to get to that any time soon, so I'm leaving my WIP here in case anybody would like to take over.

The only concern I have about this implementation is that I'm not inserting `nulls` per @nevir's guidance:

> When writing fields, any union field present in the payload should be written normally (except if not present, nulls shouldn't be inserted)

### DIFF
--- a/src/ParsedQueryNode.ts
+++ b/src/ParsedQueryNode.ts
@@ -152,8 +152,16 @@ function _buildNodeMap(
         }
       }
 
+    } else if (selection.kind === 'InlineFragment') {
+      const fragmentMap = _buildNodeMap(variables, context, fragments, selection.selectionSet, path);
+      if (fragmentMap) {
+        for (const name in fragmentMap) {
+          nodeMap[name] = _mergeNodes([...path, name], fragmentMap[name], nodeMap[name]);
+        }
+      }
+
     } else if (context.tracer.warning) {
-      context.tracer.warning(`${selection.kind} selections are not supported; query may misbehave`);
+      context.tracer.warning(`${(selection as any).kind} selections are not supported; query may misbehave`);
     }
 
     _collectDirectiveVariables(variables, selection);
--- a/test/unit/ParsedQueryNode/fragments.ts
+++ b/test/unit/ParsedQueryNode/fragments.ts
@@ -67,6 +67,31 @@ describe(`parseQuery for queries with fragments`, () => {
     });
   });
 
+  it(`parses queries with inline fragments`, () => {
+    const operation = `
+      query getThings {
+        foo {
+          ... on Foo {
+            id
+          }
+          ... on Bar {
+            id
+            name
+          }
+        }
+      }
+    `;
+    expect(parseOperation(operation)).to.deep.eq({
+      parsedQuery: {
+        foo: new ParsedQueryNode({
+          id: new ParsedQueryNode(),
+          name: new ParsedQueryNode(),
+        }),
+      },
+      variables: new Set(),
+    });
+  });
+
   it(`parses fragments with parameterized fields`, () => {
     const operation = `
       query getThings {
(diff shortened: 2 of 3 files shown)


src/ParsedQueryNode.ts
```diff
@@ -141,8 +141,14 @@ function _buildNodeMap(
         }
       }
 
-    } else {
-      context.warn(`${selection.kind} selections are not supported; query may misbehave`);
```
nevir: Let's keep the `} else {` around in case new node types are introduced down the line

src/util/tree.ts
```diff
@@ -73,9 +73,9 @@ export function walkOperation(document: DocumentNode, result: JsonObject | undef
         }
         stack.push(new OperationWalkNode(fragment.selectionSet, parent));
 
-      // TODO: Inline fragments.
-      } else {
-        throw new Error(`Unsupported GraphQL AST Node ${selection.kind}`);
```
nevir: ditto here

src/ParsedQueryNode.ts
jamesreggio: That was my original inclination, but it's not actually possible as is (!)

TypeScript determines that all possible selection types have been accounted for in the preceding cases and implicitly types `selection` as `Never` in the vestigial `else` branch. `Never` lacks a `kind` field, and thus compilation fails.

...though now I realize in my explanation that a reasonable workaround would be to just remove the `selection.kind` from the error string. Perhaps I'll just do that.

jamesreggio:
I'm fairly certain that these typing ambiguities are the reason the `IntrospectionFragmentMatcher` (and its heuristic-based friend) exist. Probably need to dig in and figure out precisely why, because there have got to be other corner cases lurking.

src/ParsedQueryNode.ts
nevir: IMO `(selection as any).kind` it instead - having the string in there will aid when debugging it

nevir:
Yeah, we could accomplish it via including the typename with these fields, and only applying them according to the typename returned

`IntrospectionFragmentMatcher` attempts to infer the "type" via the fields returned (e.g. so you can skip `__typename` auto-inclusion)

---

Somewhat related for hermes, I'd really like to push for pre-processing one's schema (similar to Relay) instead, though.  That way we can know the typenames statically as we walk payloads - and, I think, be smarter in situations like this too

jamesreggio:
Also rebased and working properly, though we really need to get the equivalent of `IntrospectionFragmentMatcher` in here eventually.

nevir:
Yeah, pulling in the schema will enable a ton of useful optimizations (like
statically knowing where entities are in the graph)

jamesreggio:
Hey @nevir — wdyt about merging this?

The code works and is (now) tested. The only deficiency is that we still write a value for fields that pertain to unmatched fragments; however, this behavior is the same for named fragments, too.

Which is to say, this already works:

```graphql
query {
  workItems {
    ...fooItem
    ...barItem
  }
}

fragment fooItem on Foo {
  bar
  bat
}

fragment barItem on Bar {
  bat
}
```

...and the PR is just making this possible:

```graphql
query {
  workItems {
    ... on Foo {
      bar
      bat
    }
    ... on Bar {
      baz
    }
  }
}
```

We should leave #153 open to track the underlying annoyance around the excessing keys/nulls, but otherwise, let's merge this so people stop saying "Hermes doesn't support unions / inline fragments."

Also, regarding the coverage issues, the uncovered portions are bits of code that are theoretically unreachable (and only there to provide friendlier errors should the GraphQL spec evolve underneath us).

nevir:
Yeah, I'm fine with merging this—

I think the one other area of confusion will be around nodes that change their type, but values from previous types will be preserved. E.g. a result like `{ bar: 1, bat: 2, baz: 3 }` could be completely possible.

PR #184, opened 2017-10-19
WIP: Add support for explicit static field via @static directive

jamesreggio:
The read-time efficiency of Hermes increases when subtrees are fully static (i.e., when `areChildrenDynamic` returns `false`). Parameterized fields and aliases are considered dynamic, and if they occur deep within the tree, nearly the entire tree must be walked to fill these values in, significantly impacting performance.

In our use case, the mappings of field alias to dynamic field arguments is always stable. This is best illustrated by an example:

```graphql
fragment EpisodeDetails on Episode {
  id
  title
  author { ...UserDetails }
  myUpvotes: upvotes(owner: ME) { totalCount }
  allUpvotes: upvotes(owner: ANY) { totalCount }
}
```

In the above example, `myUpvotes` will **always** map to the `upvotes` field with the `owner: ME` argument. (The same goes for `allUpvotes`, but with a different argument.)

If we maintain this invariant, we can treat the aliased dynamic field the same as a static field and avoid the need to overlay it dynamically at read time. To signal this, I've introduce a client-only directive called `@static`.

To use the directive, you'd add it to the aliased dynamic field like such:

```
fragment EpisodeDetails on Episode {
  ...
  myUpvotes: upvotes(owner: ME) @static { totalCount }
  allUpvotes: upvotes(owner: ANY) @static { totalCount }
}
```

In a very unscientific benchmark, I've witnessed read-time materialization speeds increase by 20% on trees with deeply-nested `@static` fields.

More work remains to flesh this out as a feature. Specifically, we need to:

* [ ] Decide whether `@static` is the right name
* [ ] Decide whether these semantics have any obscure edge cases
* [ ] Write tests
* [ ] Add dev-mode warnings if the underlying field name or arguments change for a `@static`

Comments and contributions are welcome.

### DIFF
--- a/src/ParsedQueryNode.ts
+++ b/src/ParsedQueryNode.ts
@@ -39,6 +39,10 @@ export class ParsedQueryNode<TArgTypes = JsonScalar> {
      * ignore whole subtrees in some situations if they were completely static.
      * */
     public hasParameterizedChildren?: true,
+    /**
+     * Whether the field is explicitly declared static via directive.
+     */
+    public isStatic?: true,
   ) {}
 }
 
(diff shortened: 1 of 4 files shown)


jamesreggio:
Also, because this is a client-only directive, we'll need to strip it from the document before it gets transmitted to the server. @jbaxleyiii is working on a better API for this in https://github.com/apollographql/apollo-client/issues/2309, but for now — if you'd like to try this out — you can just add the follow link to your link chain:

```javascript
import {ApolloLink} from 'apollo-link';
import {removeDirectivesFromDocument} from 'apollo-utilities';

const stripStaticDirectiveLink = new ApolloLink((operation, forward) => {
  operation.query = removeDirectivesFromDocument(
    [{name: 'static'}],
    operation.query,
  );
  return forward(operation);
});
```

jbaxleyiii:
Changes coming in the RC for the API https://github.com/apollographql/apollo-client/pull/2345

jamesreggio:
Just FYI, I've been using this in our app for a little while and it works quite nicely. Still need to finish the work to productionize, but would be great for others to take it for a spin and let me know if you run into issues.

nevir:
👍 merged this via #272 - we're going to start using this too!

PR #186, opened 2017-10-19
entityUpdaters

nevir:
update helpers on a per-typename basis, declared globally.  This allows you to guarantee that derived fields are kept in sync, even across queries.


src/CacheTransaction.ts
```diff
@@ -95,6 +104,41 @@ export class CacheTransaction implements Queryable {
...
+    const nodesToEmit = [];
+    for (const nodeId of this._editedNodeIds) {
+      const node = nextSnapshot.getNodeSnapshot(nodeId);
+      const previous = prevSnapshot.getNodeSnapshot(nodeId);
+      // One of them may be undefined; but we are guaranteed that both represent
+      // the same entity.
+      const either = node || previous;
+
+      if (!(either instanceof EntitySnapshot)) continue; // Only entities
+      const typeName = isObject(either.data) && either.data.__typename as string | undefined;
```
yuit: Should this be typename or "Query" for the rootQuery ?

src/CacheTransaction.ts
```diff
@@ -95,6 +104,41 @@ export class CacheTransaction implements Queryable {
...
+      if (!updater) continue;
+
+      nodesToEmit.push([updater, node && node.data, prevSnapshot.getNodeData(nodeId)]);
+    }
+
+    if (!nodesToEmit.length) return;
+
+    // TODO: This is weirdly the only place where we assume an Apollo interface.
+    // Can we clean this up? :(
+    const dataProxy = new ApolloTransaction(this);
```
yuit: Why do we have to create ApolloTrasaction here?  I thought we are already in one

src/CacheTransaction.ts
nevir: Doh, yeah - should be `Query` for the root node

src/CacheTransaction.ts
nevir: Nope, we're in a `CacheTransaction` :(

nevir: But in order to keep the interface consistent, gotta expose it as an apollo transaction to the callbacks.

This needs some fixing, but I think that's going to require a fair bit of refactoring to the public interfaces (as well as Apollo's own interfaces)

PR #187, opened 2017-10-20
Implement extract helper function

yuit:


src/operations/extract.ts
```diff
@@ -1,6 +1,272 @@
...
+  }
+
+  let topLevelOutboundTree: OutboundTree | [OutboundTree] | undefined;
+  let currentTreeNode: OutboundTree | [OutboundTree];
+
+  for (const outboundRef of entity.outbound) {
+    const outboundPath = outboundRef.path;
+    if (outboundPath.length > 0) {
+      // entity.data can be array if
+      // the entity is ParameterizedValueNodeSnapshot
```
nevir: `entity.data` can be _any_ value (scalars too!) for `ParameterizedValueSnapshot`s!  e.g.

```graphql
{ foo(bar:"baz") }
```

and a payload of 

```js
{ foo: 123 }
```

src/operations/extract.ts
```diff
@@ -1,6 +1,272 @@
...
+      topLevelOutboundTree = topLevelOutboundTree
+        ? topLevelOutboundTree
+        : entity.data instanceof Array ? [] : Object.create(null);
```
nevir: Can you skip the cloning for this case entirely?  (what is it protecting against?)

src/operations/extract.ts
```diff
@@ -1,6 +1,272 @@
...
+      currentTreeNode = topLevelOutboundTree!;
+    } else {
+      // path === [] then the entire data is a references
+      // stop right here.
+      // TODO (yuisu): We should assert that outboud array is of length 1
```
nevir: Nah, I can see cases where we'll have other kinds of references (such as the new nodes on the whiteboard) that would break that assertion

src/operations/extract.ts
```diff
@@ -1,6 +1,272 @@
...
+            currentTreeNode[currentPath] = null;
+          } else if (typeof outboundPath[i+1] === 'number') {
+            currentTreeNode[currentPath] = [];
+          } else {
+            currentTreeNode[currentPath] = {};
+          }
+        }
+        currentTreeNode = currentTreeNode[currentPath] as OutboundTree;
+      }
+    }
```
nevir: I think this would be a lot easier to manage & test if we had an isolated (functional) helper for removing values at a path.  It would have a lot of similarity with [`lazyImmutableDeepSet`](https://github.com/convoyinc/apollo-cache-hermes/blob/master/src/util/collection.ts#L29-L34), too (in fact, you may be able to use that directly by deep setting `undefined` and letting `JSON.stringify` strip those properties).  e.g.

```ts
function lazyImmutableDeepRemove(target, original, path) { … }
```

src/operations/extract.ts
```diff
@@ -1,6 +1,272 @@
...
+      break;
+    }
+
+    for (let i = 0; i < outboundPath.length; ++i) {
+      const currentPath = outboundPath[i];
+      if (_.isArray(currentTreeNode)) {
+        // If the currentPath is an number then there should already be an array
+        // Only create a object tree node if we still have more paths to explore
+        if (currentTreeNode.length === 0 && outboundPath[i+1] !== undefined) {
+          currentTreeNode.push({});
```
nevir: We can't guarantee that the array will contain objects though - GraphQL supports multidimensional arrays

src/operations/extract.ts
```diff
@@ -1,6 +1,272 @@
...
+    }
+
+    // ParameterizedValueNodeSnapshot can have array as top-level data value.
+    if (data instanceof Array) {
+      // For array outboundTree should only need to store
+      // one node per an entire array.
+      const result = new Array(data.length);
+      for (let i = 0; i < data.length; ++i) {
+        const element = data[i];
+        result[i] = element
```
nevir: Can we get away with only cloning this array when we need to?  I'm a little worried about potentially cloning the entire cache snapshot every time we want to serialize (that's a lot of overhead!)

src/operations/extract.ts
```diff
@@ -1,6 +1,272 @@
...
+
+    // Data is an object type
+    const output = {};
+    for (const dataName of Object.getOwnPropertyNames(data)) {
+      const outboundTreeNode = (outboundTree as OutboundTree)[dataName];
+      // Doesn't exist in outboundTree, it is a value
+      if (outboundTreeNode === undefined) {
+        reportUnSerializableError(data[dataName]);
+        output[dataName] = data[dataName];
+      } else if (outboundTreeNode === null) {
```
nevir: We need to preserve `null`s in the serialized output - they should always be explicit `null` values that came from the server

src/operations/extract.ts
```diff
@@ -1,6 +1,272 @@
...
+ *    four: '4'
+ *  }
+ * }
+ *
+ * 2)
+ * OutboundTree -> {
+ *  one: {
+ *    two: [
+ *      { three: null }
+ *    ],
```
nevir: I don't think this is safe - we can't assume that _all_ entries in the array are entities.  some may have an `id` field, whereas others may not (optional, or heterogeneous types)

I'm pretty sure we need to generate an entry in the tree for _each_ entity referenced

src/operations/extract.ts
nevir: I think you can leverage [`lazyImmutableDeepSet`](https://github.com/convoyinc/apollo-cache-hermes/blob/master/src/util/collection.ts#L29-L34) here–set each path to `undefined` (and rely on `JSON.stringify`'s behavior of removing `undefined` properties)

nevir: Thinking about it even more, that might let you _vastly_ simplify this method:

1. Collect all paths that need to be removed
2. Call `lazyImmutableDeepSet(…, …, path, undefined)` for each path

Done!

src/schema.ts
```diff
@@ -103,4 +107,26 @@ export namespace Serializeable {
...
+      for (const propName of Object.getOwnPropertyNames(value)) {
+        if (!isSerializable(value[propName])) {
+          return false;
+        }
+      }
+      return true;
+    }
+
+    if (_.isArray(value)) {
+      return value.every(element => isSerializable(element));
```
nevir: Let's avoid ES6 methods, since TS won't polyfill this - let's do regular iteration here

test/unit/operations/extract/basic/nestedReferenceWithPeerReferences.ts
```diff
@@ -0,0 +1,62 @@
...
+                four { id }
+              }
+            }
+        }`,
+      ).snapshot;
+
+      extractResult = extract(snapshot);
+    });
+
+    it(`extracts JSON serialization object`, () => {
```
nevir: grammar nit: "extracts the snapshot as a JSON-serializable object"

src/operations/extract.ts
```diff
@@ -1,6 +1,92 @@
+import * as _ from 'lodash'; // eslint-disable-line import/no-extraneous-dependencies, import/no-unresolved
```
nevir: You can remove the `import/no-unresolved` now (fixed in eslint) (https://circleci.com/gh/convoyinc/apollo-cache-hermes/2071)

src/operations/extract.ts
```diff
@@ -1,6 +1,92 @@
...
+    const entity = entities[id];
+
+    let type: Serializable.NodeSnapshotType;
+    if (entity instanceof EntitySnapshot) {
+      type = Serializable.NodeSnapshotType.EntitySnapshot;
+    } else if (entity instanceof ParameterizedValueSnapshot) {
+      type = Serializable.NodeSnapshotType.ParameterizedValueSnapshot;
+    } else {
+      cacheContext.error(`${entity.constructor.name} does not have corresponding enum value in Serializable.NodeSnapshotType`);
+      type = Serializable.NodeSnapshotType.Invalid;
```
nevir: Hmm, I'm wondering if it would be better to throw in this case?  The caller is unlikely to be able to recover from this (the serialized state will be potentially missing state)

src/operations/extract.ts
```diff
@@ -1,6 +1,92 @@
...
+    }
+
+    result[id] = serializedEntity;
+  }
+
+  return result;
+}
+
+function extractSerializableData(entity: NodeSnapshot, id: NodeId): Serializable.StringifyReadyType {
+  // If there is no outbound, then data is a value
```
nevir: Might want to mention something about the no data case, too (and why it's safe to skip, even when there are outbound fields)

src/operations/extract.ts
```diff
@@ -1,6 +1,92 @@
...
+  if (!entity.outbound || !entity.data) {
+    return entity.data;
+  }
+
+  // Shallow clone the original data
+  let extractedData: JsonValue | undefined = _.clone(entity.data);
```
nevir: Skip the clone; if all outbound fields are parameterized, it's unnecessary. 

`lazyImmutableDeepSet` will take care of it for you as it needs to:

```ts 
let extractedData: JsonValue | undefined;
for (const outbound of entity.outbound) {
  …
  extractedData = lazyImmutableDeepSet(extractedData, …)
}
```

src/operations/extract.ts
```diff
@@ -1,6 +1,92 @@
...
+      // the data at the ROOTQUERY node will be undefined with outbound
+      // reference to the parameterized node.
+      extractedData = null;
+      break;
+    }
+    // We have to check if the value exists at the path because
+    // in the case of parametrized field, the value will not exist
+    // even though there is outbound reference.
+    // So we didn't end up set the value to be 'undefined' in the output
+    if (_.get(entity.data, outbound.path)) {
```
nevir: I think it'd be a lot clearer (and resilient to future node types) if we inspect the type of node that `outbound` references (e.g. only if it is `instanceof EntitySnapshot`

Relying on `EntitySnapshot` values to be truthy may not hold for very long

test/unit/operations/extract/basic/nestedReferencesInArray.ts
```diff
@@ -25,32 +25,32 @@ describe.skip(`operations.extract`, () => {
...
-          type: Serializeable.NodeSnapshotType.EntitySnapshot,
+          type: Serializable.NodeSnapshotType.EntitySnapshot,
           outbound: [
             { id: '0', path: ['one', 'two', 0, 'three'] },
             { id: '1', path: ['one', 'two', 1, 'three'] },
           ],
           data: {
             one: {
-              two: [{}, {}],
+              two: [{ three: undefined }, { three: undefined }],
```
yuit: This is intended as we assume JSON.stringify to remove 'undefined'

src/operations/extract.ts
```diff
@@ -1,6 +1,92 @@
...
+  for (const id in entities) {
+    const entity = entities[id];
+
+    let type: Serializable.NodeSnapshotType;
+    if (entity instanceof EntitySnapshot) {
+      type = Serializable.NodeSnapshotType.EntitySnapshot;
+    } else if (entity instanceof ParameterizedValueSnapshot) {
+      type = Serializable.NodeSnapshotType.ParameterizedValueSnapshot;
+    } else {
+      cacheContext.error(`${entity.constructor.name} does not have corresponding enum value in Serializable.NodeSnapshotType`);
```
yuit: Is it correct to assume that cacheCOntext.error in our production will hook up to Sentry or other error reporting mechanism ?

src/schema.ts
```diff
@@ -71,27 +74,23 @@ export interface QuerySnapshot {
...
-    [key: string]: {
-      type: Serializeable.NodeSnapshotType,
-      inbound?: Serializeable.NodeReference[],
-      outbound?: Serializeable.NodeReference[],
-      data?: JsonValue,
-    };
+    [key: string]: Serializable.NodeSnapshot;
+  }
+
+  export type StringifyReadyType = NestedValue<JsonValue | undefined>;
```
nevir: IMO just use `JsonObject` directly; not sure we need the indirection here

src/schema.ts
```diff
@@ -100,7 +99,35 @@ export namespace Serializeable {
    *
    */
   export const enum NodeSnapshotType {
+    Invalid = -1,
     EntitySnapshot = 0,
     ParameterizedValueSnapshot = 1,
   }
+
+  export function isSerializable(value: any): boolean {
```
nevir: Might want to `value is JsonValue` this for moar bettar typing

test/helpers/context.ts
```diff
@@ -27,3 +27,6 @@ export const silentConfig: CacheContext.Configuration = {
     groupEnd: jest.fn(),
   },
 };
+
+/** Cache context created using strictConfig */
+export const strictCacheContext = new CacheContext(strictConfig);
```
nevir: 👎  we don't want a global context; that'll cache a bunch of state (queries) across test suites - will lead to surprising results

test/helpers/serialization.ts
```diff
@@ -0,0 +1,39 @@
...
+  payload: JsonObject,
+  gqlString: string,
+  cacheContext: CacheContext = strictCacheContext
+): GraphSnapshot {
+  return createSnapshot(
+    payload,
+    gqlString,
+    /* gqlVariables */ undefined,
+    /* rootId */ undefined,
+    cacheContext
```
nevir: Style: Might want to make `cacheContext` the first argument for this function, so that `variables` and `rootId` are optional and can be omitted

test/unit/operations/extract/basic/multidimensionalReferencesArray.ts
```diff
@@ -0,0 +1,84 @@
+import { extract } from '../../../../../src/operations/extract';
+import { Serializable, StaticNodeId } from '../../../../../src/schema';
+import { createOriginalGraphSnapshot, strictCacheContext } from '../../../../helpers';
+
+const { QueryRoot: QueryRootId } = StaticNodeId;
+
+describe(`operations.extract`, () => {
+  describe(`new array of values hanging off of a root`, () => {
```
nevir: "2d array of values…

test/unit/operations/extract/basic/multidimensionalReferencesArray.ts
```diff
@@ -0,0 +1,84 @@
...
+          data: { id: 'c', value: 3 },
+        },
+        'd': {
+          type: Serializable.NodeSnapshotType.EntitySnapshot,
+          inbound: [{ id: QueryRootId, path: ['rows', 1, 1] }],
+          data: { id: 'd', value: 4 },
+        },
+
+      });
+    });
```
nevir: Let's also test the case where the arrays have `null`s in them (probably as an end-to-end test that extracts & then restores)

test/unit/operations/extract/basic/nestedInvalidValues.ts
```diff
@@ -0,0 +1,33 @@
...
+            func
+          }
+        }`
+      );
+    });
+
+    it(`throws error when extracting invalid values`, () => {
+      expect(() => {
+        extract(snapshot, strictCacheContext);
+      }).to.throw(/unserializable/);
```
nevir: nit: make the regex case insensitive to make the test a bit more lenient

src/operations/extract.ts
```diff
@@ -1,6 +1,92 @@
...
+  // Set all the outbound path (e.g reference) to undefined.
+  for (const outbound of entity.outbound) {
+    if (outbound.path.length === 0) {
+      // we have to write out 'null' here to differentiate between
+      // data doesn't exist and data is a reference.
+      //
+      // In the case of parameterized field hanging off of a root
+      // the data at the ROOTQUERY node will be undefined with outbound
+      // reference to the parameterized node.
+      extractedData = null;
```
nevir: I'm wondering if we should move this case into `lazyImmutableDeepSet`?

test/unit/operations/extract/basic/multidimensionalReferencesArray.ts
yuit: I will add another test for that case here. But 👍  on end-to-end test

src/operations/extract.ts
yuit: Or may be we return undefined to indicate some failure happen? I am concerned that by throwing error here then we may get app crash etc or other behavior (is this a legitimate concern ? )

nevir: Well, it's legitimately an exceptional case; if app crashes are a concern, the caller will probably want to wrap the call in a `try` `catch` regardless to protect against any bugs

src/operations/extract.ts
yuit: `lazyImmutable...` is very powerful. it is cleaner to to just do `let extractedData = entity.data` to indicate that if no modification happen we just use the real data

src/schema.ts
yuit: I think it may be more confusing though caz when JsonValue is serializable already but this function is to check if it is serializable

src/operations/extract.ts
yuit: that will actually be better :) 🍪

src/schema.ts
yuit: The reason I have to do this because JsonObject doesn't contain 'undefined' which is desirable so I have to make another type

test/helpers/context.ts
yuit: AH!

PR #193, opened 2017-10-23
Add write test for multidimensional array

yuit:
Add write test for multi dimensional array

### DIFF
--- a/test/unit/operations/write/basic/multidimensionalReferencesArray.ts
+++ b/test/unit/operations/write/basic/multidimensionalReferencesArray.ts
@@ -1,5 +1,4 @@
 import { GraphSnapshot } from '../../../../../src/GraphSnapshot';
-import { EntitySnapshot } from '../../../../../src/nodes';
 import { NodeId, StaticNodeId } from '../../../../../src/schema';
 import { createSnapshot } from '../../../../helpers';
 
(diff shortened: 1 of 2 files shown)


test/unit/operations/write/basic/multidimensionalReferencesArray.ts
```diff
@@ -0,0 +1,73 @@
...
+    it(`records the outbound references from the query root`, () => {
+      expect(snapshot.getNodeSnapshot(QueryRootId)!.outbound).to.have.deep.members([
+        { id: 'a', path: ['rows', 0, 0] },
+        { id: 'b', path: ['rows', 0, 1] },
+        { id: 'c', path: ['rows', 1, 0] },
+        { id: 'd', path: ['rows', 1, 1] },
+      ]);
+    });
+
+    it(`directly reference each row from the query root`, () => {
```
nevir: This test is checking the shape of the objects, not what they reference.  IMO:

```ts
const rows = snapshot.snapshot.getNodeData(QueryRootId).rows;
expect(rows[0][0]).to.eq(snapshot.getNodeData('a'));
…
```

yuit: Good point!~

PR #198, opened 2017-10-24
Only set --maxWorkers when we are NOT debugging

yuit:
This is to set maximum number of workers when running the unit test. I was doing this because I keep running into out-of-memory error. But `--maxWorkers` and `--runInBands` don't work together so we have to check if we are in debugging or not

### DIFF
--- a/scripts/test-unit.sh
+++ b/scripts/test-unit.sh
@@ -23,17 +23,28 @@ for i in "${!FILES[@]}"; do
 done
 
 OPTIONS=(
-  --maxWorkers=2 --config ./test/unit/jest.json
+  --config ./test/unit/jest.json
 )
+DEBUGGING=false
+
 # Jest doesn't handle debugger flags directly.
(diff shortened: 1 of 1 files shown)


scripts/test-unit.sh
```diff
@@ -23,18 +23,29 @@ for i in "${!FILES[@]}"; do
...
+  --config ./test/unit/jest.json
 )
+DEBUGGING=false
+
 # Jest doesn't handle debugger flags directly.
 NODE_OPTIONS=()
 for option in "${OPTIONS_FLAGS[@]}"; do
-  if [[ "${option}" =~ ^--(inspect|debug-brk|nolazy) ]]; then
+  if [[ "${option}" =~ ^--(debug-brk) ]]; then
+    DEBUGGING=true
```
nevir: `inspect` implies debugging too (`debug-brk` just sets a breakpoint at launch)

PR #199, opened 2017-10-24
[Serialization] Restore function implementation

yuit:
Implement `restore` helper function. This function is tested with the `restore` tests (#197) in another PR but it is pull out into a separate one for ease of reviewing.

The PR should be checked after #197


src/operations/restore.ts
```diff
@@ -1,7 +1,92 @@
...
 
+/**
+ * Restore GraphSnapshot from serializable representation.
+ *
+ * The parameter 'serializedState' is likely to be result running JSON.stringify
+ * on a result of 'extract' method. This function will directly reference object
+ * in the serializedState.
+ *
+ * @throws Will throw an error if 'type' in serializedState cannot be mapped to
+ *    different sub-class of NodeSnapshot.
```
nevir: 👍

src/operations/restore.ts
```diff
@@ -1,7 +1,92 @@
...
+        nodeSnapshot = new EntitySnapshot(data as JsonObject, inbound, outbound);
+        break;
+      case Serializable.NodeSnapshotType.ParameterizedValueSnapshot:
+        nodeSnapshot = new ParameterizedValueSnapshot(data as JsonValue, inbound, outbound);
+        break;
+      default:
+        throw new Error(`Invalid Serializable.NodeSnapshotType ${type} at ${nodeId}`);
+    }
+
+    graphSnapshot._values[nodeId] = nodeSnapshot!;
```
nevir: `GraphSnapshot` has a constructor [for this case](https://github.com/convoyinc/apollo-cache-hermes/blob/master/src/GraphSnapshot.ts#L23); let's use that rather than exposing private stuff

src/operations/restore.ts
```diff
@@ -1,7 +1,92 @@
...
+  }
+
+  // Patch data property and reconstruct references
```
nevir: Let's break this out into a separate function for clarity

src/operations/restore.ts
```diff
@@ -1,7 +1,92 @@
+import * as _ from 'lodash';  // eslint-disable-line import/no-extraneous-dependencies
```
nevir: 👎 `lodash` is a devDependency; this'll break the package for external use (hence why `no-extraneous-dependencies` is complaining at this import)

Also, prefer the individual lodash.* packages to cut down the size of our dependency graph

src/operations/restore.ts
```diff
@@ -1,7 +1,92 @@
...
+    // If it is 'undefined' means that there is no data value
+    // in both cases, there is no need for modification.
+    if (!outbound || data === undefined) {
+      continue;
+    }
+
+    for (const { id: referenceId, path } of outbound) {
+      const referenceNode = graphSnapshot._values[referenceId];
+      if (path.length === 0) {
+        // data itsels is a reference.
```
nevir: itself - though probably clearer to just say "data is a reference"

src/operations/restore.ts
```diff
@@ -1,7 +1,92 @@
...
+        graphSnapshot._values[nodeId].data = referenceNode.data;
```
nevir: This is true for `EntitySnapshot`s, but we should not be setting `data` if the node is a `ParameterizedValueSnapshot`

src/operations/restore.ts
```diff
@@ -1,7 +1,92 @@
...
+      } else if (referenceNode instanceof ParameterizedValueSnapshot) {
+        // This is specifically to handle a sparse array which happen
+        // when each element in the array reference data in a
+        // ParameterizedValueSnapshot.
+        // (see: parameterizedFields/nestedParameterizedReferenceInArray.ts)
+        // When we do extraction of sparse array, we will represent
+        // each hole in the array as null.
+        // We will remove null to re-create a sparse array.
+        const indexToArrayIndex = _.findIndex(path, isNumber);
```
nevir: There could be multiple array indexes in the path; I don't think this is safe

src/operations/restore.ts
```diff
@@ -1,7 +1,92 @@
...
+          const arrayData = _.get(data, pathToArrayData);
+
+          if (Array.isArray(arrayData)) {
+            delete arrayData[path[indexToArrayIndex]];
+          } else {
+            // We report an error if we receive incorrect serialized result
+            // as we expect an array at this path.
+            cacheContext.error(`Expect an array at following path [${pathToArrayData}] from serialized object at nodeID ${nodeId}`);
+          }
+        }
```
nevir: I think you could simplify all this by always performing a `_.set(data, path, undefined)` (albeit maybe a bit slower)

yuit: Though that means it will no longer be sparse array -> are we ok with that ? I mean it is slightly different when iterating sparse array and non-sparse array

nevir:
Yeah, I think that's desirable.  Ideally, we should never write sparse
arrays to the cache (only support them in payloads)

src/operations/restore.ts
nevir: Yeah, I think that's desirable.  Ideally, we should never write sparse arrays to the cache (only support them in payloads)

src/GraphSnapshot.ts
```diff
@@ -2,6 +2,7 @@ import { NodeSnapshot } from './nodes';
 import { QueryResult, QueryResultWithNodeIds } from './operations/read';
 import { NodeId, OperationInstance } from './schema';
 
+export type GraphSnapshotNodesMap = { [Key in NodeId]: NodeSnapshot; };
```
nevir: Hmm, confusing name - `NodeSnapshotMap`?

nevir: Hmm, confusing name - `NodeSnapshotMap`?

src/operations/restore.ts
```diff
@@ -1,7 +1,96 @@
...
+}
+
+function createGraphSnapshotNodes(serializedState: Serializable.GraphSnapshot, cacheContext: CacheContext): GraphSnapshotNodesMap {
+  const nodesMap: GraphSnapshotNodesMap = Object.create(null);
+
+  // Create entity nodes in the GraphSnapshot
+  for (const nodeId in serializedState) {
+    const { type, data, inbound, outbound } = serializedState[nodeId];
+
+    let nodeSnapshot: NodeSnapshot | undefined;
```
nevir: It can't be `undefined` any more, now that you're throwing.  Also, I don't think you need the explicit type here - TS should infer this to `EntitySnapshot | ParameterizedValueSnapshot`

src/operations/restore.ts
```diff
@@ -1,7 +1,96 @@
...
+ * The parameter 'serializedState' is likely to be result running JSON.stringify
+ * on a result of 'extract' method. This function will directly reference object
+ * in the serializedState.
+ *
+ * @throws Will throw an error if 'type' in serializedState cannot be mapped to
+ *    different sub-class of NodeSnapshot.
+ */
 export function restore(serializedState: Serializable.GraphSnapshot, cacheContext: CacheContext): GraphSnapshot {
-  return new GraphSnapshot();
+  const _values = createGraphSnapshotNodes(serializedState, cacheContext);
```
nevir: Do you mean `values`?

src/operations/restore.ts
```diff
@@ -1,7 +1,96 @@
...
+  restoreEntityReferences(nodesMap, cacheContext);
+
+  return nodesMap;
+}
+
+function restoreEntityReferences(nodesMap: GraphSnapshotNodesMap, cacheContext: CacheContext) {
+  const { entityTransformer, entityIdForValue } = cacheContext;
+
+  for (const nodeId in nodesMap) {
+
```
nevir: style nit: kill this blank line

src/operations/restore.ts
```diff
@@ -1,7 +1,96 @@
...
+    // If it is 'undefined' means that there is no data value
+    // in both cases, there is no need for modification.
+    if (!outbound || data === undefined) {
+      continue;
+    }
+
+    for (const { id: referenceId, path } of outbound) {
+      const referenceNode = nodesMap[referenceId];
+      if (data === null) {
+        // data itsels is a reference.
```
nevir: s/itsels/itself/ - though probably clearer to just say "data is a reference"

src/operations/restore.ts
```diff
@@ -1,7 +1,96 @@
...
+        nodesMap[nodeId].data = referenceNode.data;
```
nevir: This is true for `EntitySnapshot`s, but we should not be setting `data` if the node is a `ParameterizedValueSnapshot`

src/operations/restore.ts
```diff
@@ -1,7 +1,96 @@
...
+      } else if (referenceNode instanceof ParameterizedValueSnapshot) {
+        // This is specifically to handle a sparse array which happen
+        // when each element in the array reference data in a
+        // ParameterizedValueSnapshot.
+        // (see: parameterizedFields/nestedParameterizedReferenceInArray.ts)
+        // When we do extraction of sparse array, we will represent
+        // each hole in the array as null.
+        // We will remove null to re-create a sparse array.
+        const indexToArrayIndex = lodashFindIndex(path, isNumber);
```
nevir: There could be multiple array indexes in the path; I don't think this is safe

src/operations/restore.ts
```diff
@@ -1,7 +1,96 @@
...
+        if (indexToArrayIndex !== -1) {
+          // Get the reference to an array
+          const pathToSetValue = path.slice(0, indexToArrayIndex + 1);
+
+          lodashSet(data as object, pathToSetValue, undefined);
+        }
```
nevir: I _think_ you can deep set always here

src/operations/restore.ts
yuit: you are right...though I can't just use lodashSet here since in the following case 

snipped of the extract result where `three` is a parameterized field
```
        [QueryRootId]: {
          type: Serializable.NodeSnapshotType.EntitySnapshot,
          outbound: [
            { id: parameterizedId0, path: ['one', 0, 'two', 0, 'three'] },
            { id: parameterizedId1, path: ['one', 0, 'two', 1, 'three'] },
          ],
          data: {
            one: [
              {
                four: 'FOUR',
                two: [undefined, undefined, null],
              },
            ],
          },
        },
```

we want to restore to the graph
```
         "ROOT_QUERY": EntitySnapshot {
           "data": Object {
             "one": Array [
               Object {
                 "four": "FOUR",
                 "two": Array [
                   undefined,
                   undefined,
                   null,
                ],
               },
```
Talk offline, we will have to write helper function that actually walk and set any null -> undefined as well as check to make sure that all field in data is in correct type (`array`, `object`, `not undefined`

src/operations/restore.ts
```diff
@@ -1,7 +1,128 @@
...
+ *
+ * The parameter 'serializedState' is likely to be result running JSON.stringify
+ * on a result of 'extract' method. This function will directly reference object
+ * in the serializedState.
+ *
+ * @throws Will throw an error if 'type' in serializedState cannot be mapped to
+ *    different sub-class of NodeSnapshot.
+ * @throws Will throw an error if there is undefined in sparse array
+ */
 export function restore(serializedState: Serializable.GraphSnapshot, cacheContext: CacheContext): GraphSnapshot {
```
yuit: @nevir  I think this should return `{ cacheSnapshot : new CacheSnapshot(restoreResult, restoreResult, new OptimisticUpdateQueue()) , editNodeIds} ` so that the caller can set the Snapshot through `_setSnapshot` and onChange will get called appropriately

PR #209, opened 2017-10-25
ApolloQuerable extends ApolloCache directly instead of implements DataProxy

yuit:
While trying to understanding the relationship between `Hermes`, `ApolloTransaction` and `ApolloQuerable`, I think we can have `ApolloQueryable extends from `ApolloCache` directly and implements specify methods such as diff, read etc.

Picture of my understanding -> let me know if I miss anything here
hermes_inheritance

### DIFF
--- a/src/apollo/Hermes.ts
+++ b/src/apollo/Hermes.ts
@@ -15,7 +15,7 @@ import { toQuery } from './util';
 /**
  * Apollo-specific interface to the cache.
  */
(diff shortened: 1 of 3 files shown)


nevir:
Diff seems fine, but I'm a bit iffy on it—only because I'm holding out hope that transactions will eventually _only_ have to implement `DataProxy`, and nothing else (since their primary use is in `update` callbacks)

yuit:
I was trying to do that (implementing only `DataProxy` for transaction as that will be much cleaner) but could quite get it to work :(

PR #211, opened 2017-10-26
[Serialization] End-to-End use Cache API

yuit:


src/Cache.ts
```diff
@@ -40,12 +40,15 @@ export class Cache implements Queryable {
...
 
-  restore(data: GraphSnapshot): Cache {
-    throw new Error('restore() is not implemented on Cache');
+  restore(data: Serializable.GraphSnapshot) {
+    this._snapshot.baseline = restore(data, this._context);
   }
 
-  extract() {
-    throw new Error('extract() is not implemented on Cache');
+  extract(optimistic: boolean): Serializable.GraphSnapshot {
```
nevir: Is this part of the upstream API?  I'm not sure it makes sense to serialize optimistic updates at all, since there would be no way to rescind them (no longer tied to a specific request, AFAIK)

I might bug James about this while I'm here

yuit: Yep it is the upstream API https://github.com/apollographql/apollo-client/blob/0579ff2bef140ef2e0672a51bcdab401932606e6/packages/apollo-cache/src/cache.ts#L33
Could you bug James about change the API signature too :P

nevir: Got clarity from @jbaxleyiii - currently, the only use for `optimistic` is to support dumping state to the dev tools.  Since our serialized format is different from the inmemory cache, it likely won't work anyway; but it's also trivial for us to keep this in for now.  Seems worth keeping

src/Cache.ts
```diff
@@ -40,12 +40,16 @@ export class Cache implements Queryable {
     return this._context.transformDocument(document);
   }
 
-  restore(data: GraphSnapshot): Cache {
-    throw new Error('restore() is not implemented on Cache');
+  restore(data: Serializable.GraphSnapshot) {
+    const restoreResult = restore(data, this._context);
```
yuit: `restore` should acutally return `editedNodeId`

PR #215, opened 2017-10-31
use isNil instead of explicit null check

finnigantime:
I was seeing an issue where for an object in the cache (e.g. shipment with id=0), when trying to request data on that shipment that wasn't yet in the cache (e.g. `shipment.charges` where `charges` is not in the cache because it's an additional fragment that wasn't originally requested when pulling down shipment with id=0), the cache was filling in data (including an object with all fields undefined when an array was expected for charges) instead of aborting the query traversal once a property was encountered that was not in the cache. This change seemed to fix it for me. I have not wrapped my head around this code yet, so let me know if what I'm doing is dumb and the problem is elsewhere.

### DIFF
--- a/src/operations/read.ts
+++ b/src/operations/read.ts
@@ -145,7 +145,7 @@ export function _walkAndOverlayDynamicValues(
       }
 
       // Have we reached a leaf (either in the query, or in the cache)?
-      if (node.hasParameterizedChildren && node.children && child !== null) {
+      if (node.hasParameterizedChildren && node.children && !isNil(child)) {
         if (Array.isArray(child)) {
           child = [...child];
           for (let i = child.length - 1; i >= 0; i--) {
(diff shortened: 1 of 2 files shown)


yuit:
Couple stuffs from my understanding... we can't short circuit the walk there because in the parameterized field (e.g `two(id: $id)` ) the node `data` can be underfine so we use `isNil` we will not walk into the field and get the information out... this is the reason why some of the tests are failing.

Also if we return `undefined` in the case of the test file, isn't it break the graphql contract that it will return the similar shape described by the query  ?

finnigantime:
Thanks Yui. My change here isn't correct. Closing this, and per suggestion from @nevir I will try upgrading apollo-cache-hermes and apollo-client first to see if that fixes the problem I'm hitting.

PR #217, opened 2017-11-01
Add onChange callback in cache configuration

yuit:
Add onChange to cache configuration API to allow users to perform tasks such as dispatch redux action when there is a change in the cache or log the update etc.

### DIFF
--- a/src/Cache.ts
+++ b/src/Cache.ts
@@ -124,6 +124,10 @@ export class Cache implements Queryable {
     const { snapshot, editedNodeIds } = transaction.commit();
     this._setSnapshot(snapshot, editedNodeIds);
 
+    // Call the on-change callback here to notify any change
+    if (this._context.onChange && typeof this._context.onChange === 'function') {
+      this._context.onChange(this._snapshot, editedNodeIds);
+    }
     return true;
   }
 
--- a/src/Cache.ts
+++ b/src/Cache.ts
@@ -32,11 +32,7 @@ export class Cache implements Queryable {
 
   constructor(config?: CacheContext.Configuration) {
     const initialGraphSnapshot = new GraphSnapshot();
-    this._snapshot = {
-      baseline: initialGraphSnapshot,
-      optimistic: initialGraphSnapshot,
-      optimisticQueue: new OptimisticUpdateQueue(),
-    };
+    this._snapshot = new CacheSnapshot(initialGraphSnapshot, initialGraphSnapshot, new OptimisticUpdateQueue());
     this._context = new CacheContext(config);
   }
 
--- a/src/Cache.ts
+++ b/src/Cache.ts
@@ -32,7 +32,11 @@ export class Cache implements Queryable {
 
   constructor(config?: CacheContext.Configuration) {
     const initialGraphSnapshot = new GraphSnapshot();
-    this._snapshot = new CacheSnapshot(initialGraphSnapshot, initialGraphSnapshot, new OptimisticUpdateQueue());
+    this._snapshot = {
+      baseline: initialGraphSnapshot,
+      optimistic: initialGraphSnapshot,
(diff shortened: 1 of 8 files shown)


src/CacheSnapshot.ts
```diff
@@ -4,13 +4,11 @@ import { OptimisticUpdateQueue } from './OptimisticUpdateQueue';
 /**
  * Maintains an immutable, point-in-time view of the cache.
  */
-export class CacheSnapshot {
```
nevir: 👎 keep the class: It helps enforce that we _always_ construct objects with the same shape (e.g. the same set of properties, constructed in the same order).  Makes bookkeeping in the VM lighter.  It also aids in debugging by tagging these objects with their type

See https://jsperf.com/class-vs-object-literal-creation also

src/Cache.ts
```diff
@@ -120,6 +124,10 @@ export class Cache implements Queryable {
     const { snapshot, editedNodeIds } = transaction.commit();
     this._setSnapshot(snapshot, editedNodeIds);
 
+    // Call the on-change callback here to notify any change
+    if (this._context.onChange && typeof this._context.onChange === 'function') {
```
nevir: You can drop the initial check and just `if (typeof this._context.onChange === 'function') {`

nevir: Or, actually - just go with `if (this._context.onChange) {` - we already enforce that onChange is a function via types (and could do a runtime check when constructing the cache if we want to be paranoid)

src/apollo/Hermes.ts
```diff
@@ -54,4 +55,8 @@ export class Hermes extends ApolloQueryable implements ApolloCache<GraphSnapshot
     const query = toQuery(options.query, options.variables, options.rootId);
     return this._queryable.watch(query, options.callback);
   }
+
+  getCacheSnapshot(): CacheSnapshot {
```
nevir: `currentCacheSnapshot` IMO - will help communicate that it frequently changes

src/context/CacheContext.ts
```diff
@@ -18,6 +19,8 @@ export namespace CacheContext {
   export type EntityIdMapper = (node: JsonObject) => string | number | undefined;
   export type EntityTransformer = (node: JsonObject) => void;
   export type LogEmitter = (message: string, ...metadata: any[]) => void;
+  export type OnChangeCallBack = (newCacheShapshot: CacheSnapshot, editedNodeIds: Set<String>) => void;
```
nevir: nit: `ChangeCallback` (callback is one word)

src/Cache.ts
```diff
@@ -120,6 +124,10 @@ export class Cache implements Queryable {
...
+      this._context.onChange(this._snapshot, editedNodeIds);
```
nevir: We update the snapshot in other cases ([`reset()`](https://github.com/convoyinc/apollo-cache-hermes/blob/8ba74e3c7f1717c2b3ec8cd16ed3e8611b899872/src/Cache.ts#L155)).  It would probably be clearer to move this logic into `_setSnapshot`

src/Cache.ts
yuit: Oh right the `typeof` should guard... yea I was worry about user in JS lib who doesn't have types to pass garbage in

src/Cache.ts
yuit: agree... forgot about the `reset` operation

src/CacheSnapshot.ts
yuit: Talk offline.... keep the class add comment to make sure we keep it and remove spread operator

src/Cache.ts
yuit: Actually if we gonna do runtime check we should  do for others callback as well so it is much better to do when constructing

PR #229, opened 2017-11-17
Move logging into a formal tracing interface

nevir:
This allows us to instrument the cache at various points, and callers can use it for logging, performance monitoring, APM, etc.

### DIFF
--- a/src/Cache.ts
+++ b/src/Cache.ts
@@ -117,7 +117,7 @@ export class Cache implements Queryable {
     try {
       callback(transaction);
     } catch (error) {
-      this._context.error(`Rolling back transaction due to error:`, error);
+      this._context.warn(`Rolling transaction back due to error:`, error);
       return false;
     }
 
--- a/src/Cache.ts
+++ b/src/Cache.ts
@@ -171,12 +171,21 @@ export class Cache implements Queryable {
   private _setSnapshot(snapshot: CacheSnapshot, editedNodeIds: Set<NodeId>): void {
     this._snapshot = snapshot;
 
+    let tracerContext;
+    if (this._context.tracer.broadcastStart) {
+      tracerContext = this._context.tracer.broadcastStart({ snapshot, editedNodeIds })
+    }
+
     for (const observer of this._observers) {
       observer.consumeChanges(snapshot.optimistic, editedNodeIds);
     }
 
     if (this._context.onChange) {
       this._context.onChange(this._snapshot, editedNodeIds);
+    }
+
+    if (this._context.tracer.broadcastEnd) {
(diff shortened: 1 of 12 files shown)


src/context/ConsoleTracer.ts
```diff
@@ -0,0 +1,90 @@
...
+        this._logger.warn(warning);
+      }
+      this._logger.debug('new snapshot:', newSnapshot);
+      this._logger.groupEnd();
+    } else {
+      this._logger.debug(message, { payload, newSnapshot });
+    }
+  }
+
+  // eslint-disable-next-line class-methods-use-this
```
nevir: Sticking it on the class so that this tracer can be extended, and this method overridden

src/Cache.ts
```diff
@@ -117,7 +117,9 @@ export class Cache implements Queryable {
     try {
       callback(transaction);
     } catch (error) {
-      this._context.error(`Rolling back transaction due to error:`, error);
```
nevir: Converting this (And all other errors) into `warn` calls, since they don't actually _stop_ the cache from operating

src/Cache.ts
```diff
@@ -169,13 +171,22 @@ export class Cache implements Queryable {
   private _setSnapshot(snapshot: CacheSnapshot, editedNodeIds: Set<NodeId>): void {
     this._snapshot = snapshot;
 
+    let tracerContext;
+    if (this._context.tracer.broadcastStart) {
```
yuit: Q: why are we using broadcast here?

src/context/Tracer.ts
```diff
@@ -0,0 +1,80 @@
...
+
+  /**
+   * Start of a request to read from the cache.
+   */
+  readStart?: (rawOperation: RawOperation) => TActionContext;
+
+  /**
+   * Successful end of a request to read from the cache.
+   */
+  readEnd?: (operation: OperationInstance, info: Tracer.ReadInfo, context: TActionContext) => void;
```
yuit: Q: Why we separate into `readEnd` and `writeEnd` instead of having one function and then passing string of operation type?

src/Cache.ts
nevir: Similar to the current tracing we're doing with the default cache, it's helpful to see the cost of broadcasting a given query (since react Apollo turns around and immediately reads new query results)

src/context/Tracer.ts
nevir: Couple reasons:

* It makes it easier to short circuit (if you don't implement one, it'll not call it)
* The types are subitly different - and I didn't want to go crazy with overloads

PR #242, opened 2017-12-01
Fix read traversal

jamesreggio:
The test demonstrates the problem, but basically, if you have two nested, parameterized fields without any simple (non-parameterized) fields on the intermediate node, the intermediate node will be written into the snapshot with `null` for its `data`, which will short-circuit the dynamic value overlay walk.

The `nextContainerId` and `nextPath` juggling is necessary to ensure that arguments are affixed in the correct order as we build the path to deeper nodes. Without it, the next `OverlayWalkNode` would simply incorporate the `fieldName` of the parent, and its arguments would be lost.

### DIFF
--- a/src/operations/read.ts
+++ b/src/operations/read.ts
@@ -125,6 +125,9 @@ export function _walkAndOverlayDynamicValues(
       // This is an alias if we have a schemaName declared.
       fieldName = node.schemaName ? node.schemaName : key;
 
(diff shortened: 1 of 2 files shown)


nevir:
Now that I've got some benchmarks, this one unfortunately looks like it significantly impacts reads without a large amount of entity overlap (or small updates).

I think we'll need to tweak the approach somewhat

PR #243, opened 2017-12-01
Include parameterized fields in nodeIds when reading

jamesreggio:
In the interest of better tracking around read and write `nodeIds` (for the sake of avoiding unnecessary update broadcasts), I've fixed an omission where parameterized fields were not being included in the `nodeIds` during a read operation. These fields should be included because all parameterized fields are considered top-level nodes by Hermes.

Test included, works in our app.


nevir:
👍  thanks! Dealt w/ the conflicts & merged this one via #359

PR #247, opened 2017-12-05
walk ParsedQueryNode instead of documentNode in read operation

yuit:
This PR  is trying to make thing more consistent between read/write operation. We usually walk  ParsedQueryNode when doing operation except when we verify the read result.

### DIFF
--- a/src/util/tree.ts
+++ b/src/util/tree.ts
@@ -1,19 +1,9 @@
-import { // eslint-disable-line import/no-extraneous-dependencies
-  DocumentNode,
-  FieldNode,
-  SelectionSetNode,
-} from 'graphql';
-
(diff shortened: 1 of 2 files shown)


src/util/tree.ts
```diff
@@ -1,76 +1,50 @@
...
-        if (!fragment) {
-          throw new Error(`Expected fragment ${selection.name.value} to be defined`);
-        }
-        stack.push(new OperationWalkNode(fragment.selectionSet, parent));
-
-      // TODO: Inline fragments.
-      } else {
-        throw new Error(`Unsupported GraphQL AST Node ${selection.kind}`);
+    for (const fieldName in parsedOperation) {
+      const child = get(parent, fieldName);
```
nevir: nit: let's move `child` into the condition

PR #252, opened 2017-12-08
Always transformDocument before read/write Operation

yuit:


yuit:
@nevir is fixing this in Apollo! Yea! but I will merge it in to get us unblocked on our app.

PR #254, opened 2017-12-12
Specific error being thrown for partial read 

JAStanton:
Apollo cache hermes throws and error with message `read not satisfied by the cache.` upon incomplete read. As a consumer I want to identify this error as a specific error instance and ignore them. The best way to do this is check error instance of. So this change throws a new `UnsatisfiedCacheError` and exposes it as well as other errors so I can sanity check them on my app.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "apollo-cache-hermes",
-  "version": "0.4.1-alpha",
+  "version": "0.4.0-alpha",
   "description": "A cache implementation for Apollo Client, tuned for performance",
   "license": "Apache-2.0",
   "repository": "convoyinc/apollo-cache-hermes",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "apollo-cache-hermes",
-  "version": "0.4.0-alpha",
+  "version": "0.4.1-alpha",
   "description": "A cache implementation for Apollo Client, tuned for performance",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "apollo-cache-hermes",
-  "version": "0.4.0-alpha",
+  "version": "0.4.1-alpha",
```
nevir: This package auto-bumps the z value (but doesn't check it in) - leave it as `.0`

src/apollo/Queryable.ts
```diff
@@ -1,6 +1,7 @@
 import { Cache, DataProxy } from 'apollo-cache';
 import { DocumentNode } from 'graphql'; // eslint-disable-line import/no-extraneous-dependencies
 
+import { UnsatisfiedCacheError } from '../errors';
```
nevir: So unsatisfying 👍

src/errors.ts
```diff
@@ -24,6 +24,16 @@ export class QueryError extends CacheError {
...
 }
 
+/**
+ * An error with a read query - generally occurs when data in cache is partial
+ * or missing.
+ */
+export class UnsatisfiedCacheError extends CacheError {
+  constructor(message: string) {
+    super(message);
+  }
```
nevir: This constructor isn't giving us anything - drop it & one-line the definition here!

PR #255, opened 2017-12-12
Add updateListOfReferences

yuit:
**Summary**
An API in `Hermes` which allow users to update each field instance on a given container. This is useful in `EntityUpdater` (see tests for more details)

Minor fix on `EnityUpdater` with optimistic write


src/apollo/Transaction.ts
```diff
@@ -45,4 +59,45 @@ export class ApolloTransaction extends ApolloQueryable implements ApolloCache<Gr
...
+   */
+  updateFieldInstances(
+    containerId: NodeId,
+    editPath: PathPart[],
+    { fragment, fragmentName }: {fragment: DocumentNode, fragmentName?: string},
+    updateFieldCallback: (previousValues: StoreValue | undefined, fieldArgs: { [argName: string]: string }) => any
+  ) {
+    const currentContainerNode = this._queryable.getCurrentNodeSnapshot(containerId);
+    if (!currentContainerNode || !currentContainerNode.outbound) {
+      // TODO (yuisu): error handling
```
yuit: Now that I think about it. This may be ok (e.g it is not an error)

src/apollo/Transaction.ts
```diff
@@ -45,4 +59,45 @@ export class ApolloTransaction extends ApolloQueryable implements ApolloCache<Gr
...
+      return;
+    }
+
+    for (const { id: outboundId, path } of currentContainerNode.outbound) {
+      if (lodashIsEqual(editPath, path)) {
+        const fieldArguments = getOriginalFieldArguments(outboundId);
```
bienur: What about the case when the field is not parameterized?

yuit: There will be no fieldArguments so it will be doing nothing

PR #272, opened 2017-12-26
@static directive

nevir:
Continuation of #184 with the following changes:

* `@static` is now fully isolated to ParsedQueryNode (static fields are presented to read/write as part of the schema)
* Tests

For dev-mode warnings; I think we should defer until we start plumbing knowledge of the app's schema through hermes


jamesreggio:
Ah, this is a lovely simplification. Thanks for shipping this!

PR #275, opened 2018-01-02
Fix optimistic transactions

jamesreggio:
Prior to this change, if you had multiple optimistic transactions, you'd only see the latest one reflected in the optimistic snapshot. This one-line change fixes the issue and includes a test.


nevir:
Well, hell.

PR #282, opened 2018-01-09
WIP: Add support for `@include` and `@skip` directives

jamesreggio:
This generally works for us, but I've left some concerns as `TODO` below.

My main concern is that the directives on the `SelectionNode` need to be evaluated at the time that the operation's final variables are available. Attaching the `SelectionNode` to the `ParsedQueryNode` kinda defeats the purpose of having a `ParsedQueryNode`, but it works for us for now.

Going forward, I think we'll need to pluck the relevant directives (and their variable dependencies) out of the `SelectionNode` at parse time and evaluate them similar to how we evaluate arguments. The only problem with this approach is that `shouldInclude` (in `apollo-utilities`) is a bit too coarse-grained and expects to operate on a full `SelectionNode`. Obviously, we could propose a change to `apollo-utiltiies`, if needed.

No tests or anything yet.

### DIFF
--- a/src/ParsedQueryNode.ts
+++ b/src/ParsedQueryNode.ts
@@ -1,4 +1,4 @@
-import { isEqual, valueFromNode } from 'apollo-utilities';
+import { isEqual, valueFromNode, shouldInclude } from 'apollo-utilities';
 import { // eslint-disable-line import/no-extraneous-dependencies
   ArgumentNode,
   SelectionSetNode,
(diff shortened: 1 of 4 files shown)


nevir:
Some day we'll get to this, I swear

jamesreggio:
Yeah, sure...

jamesreggio:
Just so you know, I've rebased this and made it a bit more concise.

Still no tests.

PR #283, opened 2018-01-10
add cache pruning support

bienur:
Support for a new operation, `prune()`, capable of pruning the cache to the shape of a query. One of the motivation behind this is that when used in conjunction with `extract()`, client can serialize just what it needs, instead of the entire cache, to persisted media and leave all other artifacts in the cache behind.


src/operations/prune.ts
```diff
@@ -0,0 +1,21 @@
...
+/**
+ * Return a new graph snapshot pruned to just the shape of the given query
+ */
+export function prune(context: CacheContext, snapshot: GraphSnapshot, raw: RawOperation): EditedSnapshot {
+  const queryResult = read(context, raw, snapshot);
+  return write(
+    context,
+    new GraphSnapshot(),
+    raw,
+    queryResult.result && queryResult.complete ? queryResult.result : {} as JsonObject
```
yuit: Why would we not write incomplete result here? It seems unexpected that if the read result is incomplete, we will silently write {}.... I wonder iit will be best to actually log the fact that it is incomplete and just return empty `GraphSnapshot`?

src/operations/verify.ts
```diff
@@ -0,0 +1,13 @@
+import { CacheContext } from '../context';
+import { GraphSnapshot } from '../GraphSnapshot';
+import { RawOperation } from '../schema';
+
+import { read } from './read';
+
+/**
+ * Verify if a query can be satified by the given graph snapshot.
+ */
+export function verify(context: CacheContext, raw: RawOperation, snapshot: GraphSnapshot): boolean {
```
yuit: Q: What is this function for ?

bienur: This is a separate operation that will be used by the verification step post-migration.

src/operations/prune.ts
bienur: We do write and the snapshot returned is an empty `GraphShapshot` (all be it wrapped inside an `EditedShapshot`, as  that is the return type).

PR #287, opened 2018-01-12
add cache migrate support and tests

bienur:
Support for a new operation, `migrate()`, which takes a `migrationMap` and migrates entities in the cache from their old shape. Useful in upgrading the persisted data following a `restore`.

### DIFF
--- a/src/operations/migrate.ts
+++ b/src/operations/migrate.ts
@@ -5,7 +5,9 @@ import { isObject, isReferenceField } from '../util';
 
 /**
  * Function called during migration of entities to add or update a field.
- * The new value will be whatever the function evaluates to.
+ * The new value will be whatever the function evaluates to. If the field to
+ * be migrated is a nested object, the migration function should produce a new
+ * object, instead of re-using the old one, and make the changes on top.
  */
 export type FieldMigration = (previous: JsonValue) => any;
 export type MigrationMap = {
@@ -15,8 +17,8 @@ export type MigrationMap = {
 };
 
 /**
- * Return the migrated json object. Supports add and modify but not delete
- * fields
+ * Returns the migrated entity snapshot. Supports add and modify but not delete
+ * fields.
  */
 export function migrateEntity(snapshot: EntitySnapshot, migrationMap?: MigrationMap): EntitySnapshot {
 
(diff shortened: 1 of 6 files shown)


src/operations/migrate.ts
```diff
@@ -0,0 +1,56 @@
...
+ */
+export type FieldMigration = (previous: JsonValue) => any;
+export type MigrationMap = {
+  [typeName: string]: {
+    [fieldName: string]: FieldMigration,
+  },
+};
+
+/**
+ * Return the migrated json object. Supports add and modify but not delete
```
yuit: nit: it doesn't return JSON. From the signature, it returned the `EntitySnapshot`
also use JSDoc style `@return ...`

src/operations/migrate.ts
```diff
@@ -0,0 +1,56 @@
...
+ * fields
+ */
+export function migrateEntity(snapshot: EntitySnapshot, migrationMap?: MigrationMap): EntitySnapshot {
+
+  // Only if object and if valid migrationMap is provided
+  if (!isObject(snapshot.data) || !migrationMap) return snapshot;
+
+  let typeName = snapshot.data.__typename as string | undefined;
```
yuit: I don't think you need the cast here

src/operations/migrate.ts
```diff
@@ -0,0 +1,56 @@
...
+  if (!typeName) typeName = 'Query';
+  if (!migrationMap[typeName]) return snapshot;
+  for (const field in migrationMap[typeName]) {
+    const fieldMigration = migrationMap[typeName][field];
+    if (!fieldMigration) continue;
+    if (isReferenceField(snapshot, [field])) {
```
yuit: Why do we not support reference field ? -> could you also document the reason in the comment

src/operations/migrate.ts
bienur: Good catch.

Not doing @returns for now since jslint doesn't seem to like that much.

bienur: Not doing @returns for now since jslint doesn't seem to like that much.

src/operations/migrate.ts
bienur: Clarified offline

PR #288, opened 2018-01-13
Expose migration on Hermes

bienur:
This wires up the `prune` and `migration` operation PR-ed previously to `Cache` and expose migration functionality on `Hermes`' interface. With tests, of course. 😄

### DIFF
--- a/src/apollo/Hermes.ts
+++ b/src/apollo/Hermes.ts
@@ -4,7 +4,7 @@ import {
   ApolloCache,
 } from 'apollo-cache';
 
-import { Cache } from '../Cache';
+import { Cache, MigrationMap } from '../Cache';
 import { CacheSnapshot } from '../CacheSnapshot';
 import { CacheContext } from '../context';
 import { GraphSnapshot } from '../GraphSnapshot';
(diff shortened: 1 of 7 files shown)


src/apollo/Hermes.ts
```diff
@@ -26,14 +26,16 @@ export class Hermes extends ApolloQueryable implements ApolloCache<GraphSnapshot
   }
 
   // TODO (yuisu): data can be typed better with update of ApolloCache API
-  restore(data: any): ApolloCache<GraphSnapshot> {
-    this._queryable.restore(data);
+  restore(data: any, migrationMap?: MigrationMap, verifyOptions?: CacheInterface.ReadOptions): ApolloCache<GraphSnapshot> {
```
bienur: Not entirely sure about the use of Apollo’s `Cache.ReadOptions` for prune and verify queries. Is there a better type for the client to supply queries to hermes?

yuit: I think the `ReadOptions` is reasonable...It has minimal type needed... if you want it to have better name feel free to type alias too

PR #295, opened 2018-01-18
Add (broken) test for data loss when unrelated fragments have overlapping fields

jamesreggio:
Per a Slack conversation with @nevir.

This should ideally work when #153 is resolved.


jamesreggio:
I love when you merge my broken code <333

PR #297, opened 2018-01-27
Try catch stackoverflow in isSerializable of extract

yuit:
@jamesreggio  ran into an infinite recursion during extract. This is occur inside `isSerializable` function. This is possible a result of circular reference in data. Before we are trying to instrument any fix of circular reference which is not trivia. We will catch the stack overflow error to get more context


jamesreggio:
Thanks, @yuit! I've shipped this today and will let you know what I discover from the additional logging. (Also, per our Slack conversation, I confirmed that you can indeed `catch` a stack overflow in JSCore on iOS.)

PR #317, opened 2018-02-24
Fix optimistic updates not getting rolled back

TimothyKrell:
When `transaction.rollback` was called, the resulting optimistic snapshot still included the rolled-back snapshot. The reason was `transaction._buildOptimisticSnapshot` was using the current `optimisticQueue` instead of the new one generated after removing the rolled-back update from the queue. Added failing test for the bug.


nevir:
Well hell!  Thanks for catching this & adding the test 👍

PR #319, opened 2018-03-01
Implement connection directive

bennyhobart:
updated `buildFieldArgs` to now resolve correctly for the `connection` directive


bennyhobart:
@nevir

nevir:
Sorry, probably going to be a few more days before I can get around to taking a look at this one :(

bennyhobart:
all good at your own time! thanks for building the entire cache ;) also I imagine you'll want to provide feedback on how we should update the cache — right now I am just overwriting with the new result, 0 smarts.

It would be great for application developers to be able add their own "Custom directive handlers" (like (Relay Handlers)[https://github.com/facebook/relay/blob/master/packages/relay-runtime/handlers/RelayDefaultHandlerProvider.js] that contain information on how to update and read from the cache but I'm not certain how to go about implementing this if you have any ideas.

nevir:
Depending on what you're trying to achieve - you might be able to implement your directive handlers as ApolloLinks (the cache gets attached to the link context, IIRC)

nevir:
Also, I know it's not well documented yet - but take a look at Hermes' change handlers as well:

https://github.com/convoyinc/apollo-cache-hermes/blob/master/src/context/CacheContext.ts#L98-L108

hammadj:
@nevir Any update on when this can be merged?

nevir:
Apologies! I know I've severely dropped the ball on this one :(

Hesitant to give an ETA - and I've been hesitant to merge because I don't
know the @connection spec well enough yet to have a good opinion on it

Considering merging it with a cursory review.  Would you folk be ok with
helping support issues related to if they come up?

PR #331, opened 2018-04-20
Initial coding for parameterized field migration

bienur:
The code is not yet fully tested but gives the basic idea about the scenarios that we cover and the approach we take. Just want to get an early feedback on whether or not you see any fundamental issue.

### DIFF
--- a/src/operations/migrate.ts
+++ b/src/operations/migrate.ts
@@ -1,7 +1,20 @@
+// import lodashGet = require('lodash.get');
+import * as _ from 'lodash';
+import { isEqual } from 'apollo-utilities';
+
 import { CacheSnapshot } from '../CacheSnapshot';
-import { EntitySnapshot } from '../nodes';
-import { JsonValue } from '../primitive';
-import { isObject, isReferenceField } from '../util';
+import { GraphSnapshot } from '../GraphSnapshot';
+import { EntitySnapshot, ParameterizedValueSnapshot, NodeSnapshot } from '../nodes';
+import { JsonValue, PathPart } from '../primitive';
+import { NodeId } from '../schema';
+import {
+  isObject,
+  isReferenceField,
+  addNodeReference,
+  removeNodeReference
+} from '../util';
+
+import { nodeIdForParameterizedValue, NodeSnapshotMap } from './SnapshotEditor';
 
 /**
  * Function called during migration of entities to add or update a field.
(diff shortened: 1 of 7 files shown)


src/operations/migrate.ts
```diff
@@ -10,52 +23,194 @@ import { isObject, isReferenceField } from '../util';
...
+    const nodes = cacheSnapshot.baseline._values;
+    for (const nodeId in nodes) {
+      const nodeSnapshot = nodes[nodeId];
       if (nodeSnapshot instanceof EntitySnapshot) {
-        migrateEntity(nodeSnapshot, migrationMap);
+        migrateEntity(nodeId, nodeSnapshot, nodesToAdd, migrationMap);
+      } else if (nodeSnapshot instanceof ParameterizedValueSnapshot) {
+        if (shouldGarbageCollect(nodeId, cacheSnapshot.baseline, migrationMap)) {
+          makeOrphan(nodeId, cacheSnapshot.baseline, nodesToRemove);
+        }
```
bienur: For old parameterized value snapshots, we check if the parameterized field is out-of-date per the migration map. If so, we make it an orphan, then transitively collects nodes that are made orphans into `nodesToRemove`.

src/operations/migrate.ts
```diff
@@ -10,52 +23,194 @@ import { isObject, isReferenceField } from '../util';
...
+    const snapshots = { ...cacheSnapshot.baseline._values };
+    for (const removeId in nodesToRemove) {
+      delete snapshots[removeId];
+    }
+    for (const addId in nodesToAdd) {
+      const nodeToAdd = nodesToAdd[addId];
+      if (!nodeToAdd) continue;
+      snapshots[addId] = nodeToAdd;
+    }
+    cacheSnapshot.baseline = new GraphSnapshot(snapshots);
```
bienur: We rebuild the migrated Graphsnapshot before returning.

src/operations/migrate.ts
yuit: I still don't understad why we want to remove them? we don't know for sure that it won't be use. Just because we have new signature, it is not enough to say that there is no usage in other location in app

src/operations/migrate.ts
```diff
@@ -10,52 +23,194 @@ import { isObject, isReferenceField } from '../util';
...
-    for (const id in entities) {
-      const nodeSnapshot = entities[id];
+    const nodesToAdd: NodeSnapshotMap = Object.create(null);
+    const nodesToRemove: NodeSnapshotMap = Object.create(null);
+    const nodes = cacheSnapshot.baseline._values;
+    for (const nodeId in nodes) {
+      const nodeSnapshot = nodes[nodeId];
       if (nodeSnapshot instanceof EntitySnapshot) {
-        migrateEntity(nodeSnapshot, migrationMap);
+        migrateEntity(nodeId, nodeSnapshot, nodesToAdd, migrationMap);
```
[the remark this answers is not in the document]
bienur: Container node is always an entity node.

The scenario that the new migrateEntity logic trying to cover is the following:

Persisted:
Foo {
 Apple,
 Banana,
 ...
}

src/operations/migrate.ts
bienur: It won't be used elsewhere for sure because a parameterized value is designed to be an overlay for a specific entity instance (the format of the id suggests so). To be used elsewhere implies there needs to another entity that is both the same and not the same as the parameterized snapshot's container which just isn't possible.

yuit: it is possible if the user use graphql alias ( we don't use it ourselves) within same container....the id format encoding the parameterized arguments so the graphsnapshot is not limited to same graphql field name with different parameters..

test/unit/Apollo/persistenceRoundTrip.ts
```diff
@@ -116,8 +118,10 @@ describe(`extract/restore roundtrip`, () => {
 
     expect(() => {
       hermes.restore(JSON.parse(persisted), {
-        ['Viewer']: {
-          ['age']: _previous => '',
+        _entities: {
+          ['Viewer']: {
```
yuit: why do you wrap the string inside computed property ? It is already a string

bienur: There's no need. Can get rid of them all.

PR #336, opened 2018-05-23
Short-circuit repeated writes of the same entity

jamesreggio:
This commit adds a SelectionSetCache that hashes each SelectionSet and returns a shared ParsedQueryNode for identical queries.

The resulting referential equality of ParsedQueryNodes allows us to avoid writing an entity with the same ID and query selection to the store multiple times during a write transaction, which can be quite beneficial for deeply-nested query responses (which often contain a lot of duplication).

In our app's 'notifications' query (which is complex and has many nested objects, and averages 400kb uncompressed), this change dropped the write time from a typical range of 400-650ms to between 30-40ms.

As with most my PRs, this is lacking tests, though it does not regress any existing tests.

### DIFF
--- a/src/ParsedQueryNode.ts
+++ b/src/ParsedQueryNode.ts
@@ -87,6 +87,41 @@ export class VariableArgument {
 }
 
 /**
+ * Maintains a cache of SelectionSetNode to ParsedQuery mappings.
+ *
+ * The cache is keyed on the JSON-stringified SelectionSetNode, so equivalent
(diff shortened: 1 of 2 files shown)


nevir:
Now that I've got some benchmarks, this one unfortunately looks like it significantly impacts (2-3x on the bench) reads without a large amount of entity overlap (or small updates)

Will need to investigate some more; as that's mostly in anecdote territory, but I think we'll need to tweak the approach somewhat

PR #244, opened 2017-12-01
WIP: Re-enable smart updates based upon nodeIds in prior read results

jamesreggio:
With the changes in #242 and #243, I've been able to re-enable the intelligent updates originally specified in `_hasUpdates`.

We need to test this thoroughly, but I'm curious if any unaddressed edge cases immediately come to mind?

PR #362, opened 2018-07-06
Add readCache carryforward

jamesreggio:
This is another small optimization we've been using for the past 3+ months.

This carries forward cached read results during a cache transaction (i.e., during a write) if the transaction did not modify any of the `nodeIds` referenced within a cached read-result.

This logic hinged on the correctness of the `editedNodeIds` and `_result.nodeIds` sets, and these sets appear to be full-baked now, given the merger of #244 and #360.

### DIFF
--- a/src/Cache.ts
+++ b/src/Cache.ts
@@ -186,7 +186,27 @@ export class Cache implements Queryable {
    * Call onChange callback if one exist to notify cache users of any change.
    */
   private _setSnapshot(snapshot: CacheSnapshot, editedNodeIds: Set<NodeId>): void {
+    const lastSnapshot = this._snapshot;
     this._snapshot = snapshot;
+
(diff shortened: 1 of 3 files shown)


nevir:
Awesome! Merging this one via #362 

One case where this falls over is for queries where we are causing changes to the query root, which busts pretty much every query (unless it has only parameterized fields at the top level). 

We might be able to get away with special casing changes to the query root, but that requires more investigation/proof

jamesreggio:
Sweeeeet. Thanks for getting this over the line!

And yeah, I've noticed that problem with the nuke-the-world `ROOT_QUERY` invalidation. For the time being, I've just been working around it by assigning a non-static alias to non-parameterized top-level fields, but Hermes certainly could do better here.

Issue #311, opened 2018-02-16
Rolling back cache due to error: InvalidPayloadError: unsupported transition from a non-entity value to an entity


finnigantime:
**Summary**:
Encountered: `Rolling back cache due to error: InvalidPayloadError: unsupported transition from a non-entity value to an entity`.

<img width="1179" alt="36220100-54186f2e-116e-11e8-91be-1bb49e629000" src="https://user-images.githubusercontent.com/1305608/36317597-5e409992-12f2-11e8-9a89-08aa7c48e859.png">

This occurred when refetching a different subfragment of a note under a different query.

nevir:
So the error here is that you originally selected a field as a non-entity
(e.g. without an `id` field), but in a subsequent request, selected it as
an entity (with an `id`).

From the cache's point of view, this is an ambiguous update: it can't know
if the first value is the same, or a different entity (and whether it
should merge values, or replace them)

The only safe thing to do would be to replace, but that breaks assumptions
and potentially invalidates previously executed queries (if they selected
any fields not selected by the new fragment)

---

TL;DR - make sure all fragments for that entity select the `id` field

nevir:
TL;DR - make sure all fragments for that entity select the `id` field (or that none of them do; but I'm pretty sure that's not what you want in this case)

nevir:
Let's leave this issue open as a reminder that that error could use some more documentation

boorad:
I'm seeing this error as well, but not for a cache rollback.  I'm trying to writeFragment and change an entity that was a sub-entity of a larger query.  A 'Game' object (fetched with a getGame query) has multiple Round objects in an array.  I'm trying to writeFragment to one of the rounds with _key (id field), but adding the id field as suggested here does not work.  Do I need to have a getRound query for each Round object and do a writeQuery instead?  Seems like one fetch of a game and 12 rounds is more efficient.  I just want to update this fragment of 'scores' key on the round object.  A view of the cache is here:

![cache](https://user-images.githubusercontent.com/15431/42654908-40225806-85e8-11e8-803f-9ce3aa1634f9.png)

nevir:
Hmm, @boorad, the cache generally doesn't care how you reach a given entity (via query, fragment, etc); for this error, I think what is likely is that you've queried an entity once without `_key` (the game in `getGame`?), and are attempting to write a fragment that now includes a `_key` for that same path in the graph

[...]

PR #364, opened 2018-07-12
Busted test for writeFragment for an entity that doesn't inline the entity's id

nevir:
See #311 for context


nevir:
Killing off some of my really old PRs

Issue #358, opened 2018-06-28
Switch from chai to jasmine expectations


nevir:
We're currently breaking assumptions somewhat by using chai's `expect` under jest tests (which normally pulls in jasmine's `expect`).  Let's switch over to clean that up

PR #380, opened 2018-10-05
[Tests] Translate some tests to Jest expect

monotkate:
## What do!?

Adds `jestExpect` in parallel with chai `expect` (default).  Allows for the ability to translate tests piecemeal to jest.  Once complete can replace `jestExpect` with `expect`.

https://github.com/convoyinc/apollo-cache-hermes/issues/358

## Tests!

This PR adds jasmine support, and the converts all the tests under `unit/operations/restore` to Jasmine.

## Basic transitions:

- `to.eq`-> `toBe`
- `to.deep.eq` -> `toEqual`
- `to.have.deep.members` -> `toEqual` 
(from my understanding of `to.have.deep.members` it translated to `toEqual`)
- `to.be.an.instanceOf` -> `toBeInstanceOf`
- `to.throw` -> `toThrow`

### DIFF
--- a/test/env/base.ts
+++ b/test/env/base.ts
@@ -4,9 +4,14 @@ import chaiJestDiff from 'chai-jest-diff';
 
 // Chai
 
-// We prefer Chai's `expect` interface.
+// We are transitioning to using Jasmine for our test expects.  During the
+// transition, Jasmine and chai will be run in parallel using jestExpect for
+// Jasmine, and the default expect for chai.  Once complete, Jasmine will
+// replace the chai global.expect.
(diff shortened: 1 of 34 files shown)


test/env/base.ts
```diff
@@ -5,6 +5,7 @@ import chaiJestDiff from 'chai-jest-diff';
 // Chai
 
 // We prefer Chai's `expect` interface.
+global.jestExpect = global.expect;
```
nevir: Let's leave a comment saying this is temporary, and a note about the migration process

PR #384, opened 2018-10-16
[Tests] Convert Write Tests to Jasmine

monotkate:
## WhatDo!?

Convert the write tests to Jasmine.

## Basic transitions:

- `to.eq`-> `toBe`
- `to.deep.eq` -> `toEqual`
- `to.have.deep.members` -> `toEqual(expect.arrayContaining())` 
- `to.have.members` -> `toEqual(expect.arrayContaining())` 
- `to.be.an.instanceOf` -> `toBeInstanceOf`
- `to.throw` -> `toThrow`


monotkate:
I agree that `arrayContaining()` is a better approach than sorting both sides, however, `.toMatchObject` looks for number of elements and does _not_ allow you to return more data than expected.  I think `arrayContaining()` is the appropriate choice here as well.

> You can also pass an array of objects, in which case the method will return true only if each object in the received array matches (in the toMatchObject sense described above) the corresponding object in the expected array. This is useful if you want to check that two arrays match in their number of elements, as opposed to arrayContaining, which allows for extra elements in the received array.

nevir:
SGTM 👍

Issue #412, opened 2019-04-05
writeFragment blows away entity's __typename


finnigantime:
**Summary**:

Say existingFoo is an entity in the cache with this data:

```
{
  __typename: 'Foo',
  id: '123',
  bar: oldBarValue,
  baz: 'bazValue'
}
```

We call `writeFragment()` to update it's `.bar`:
```
    proxy.writeFragment({
      id: existingFoo.id,
      fragment: fragments.fooWithBar,
      fragmentName: 'fooWithBar',
      data: {
        id: existingFoo.id,
        bar: newBarValue,
      },
    });
```

**Expected Behavior**:

```
{
  __typename: 'Foo',
  id: '123',
  bar: newBarValue,
  baz: 'bazValue'
}
```

**Actual Behavior**:

```
{
  __typename: undefined,
  id: '123',
  bar: newBarValue,
  baz: 'bazValue'
}
```

This happens when `addTypename` is true in the config since the fragment we are writing gets transformed and `__typename` gets added to it. If we don't similarly include `__typename: 'Foo'` in the `data` arg of writeFragment, the existingFoo.__typename gets deleted.

This is not obvious and at the least should result in a warning. Ideally it shouldn't happen at all (when writing a subfragment, I don't think __typename should be deleted if it wasn't specified on the subfragment). Recipes for writeFragment don't indicate the need to add __typename (https://www.apollographql.com/docs/angular/basics/caching#writequery-and-writefragment) and it has caused problems for users (https://stackoverflow.com/questions/48631954/apollo-writefragment-not-updating-data).

nevir:
If the id has changed from what previously existed at that position in the
graph, this is expected (you're replacing the entity with a new one, and
Hermes can't make assumptions about fields from the previous entity - what
if it's a union type?)

nevir:
General guidance is that you'll need to provide __typename when calling
writeFragment (or any direct write to the cache) to ensure that it stays
consistent

nevir:
Ah, the alternative is that it is a subtle bug. Because we have addTypename
enabled, we are implicitly asking for __typename in the fragment passed to
the cache, and because it's not present in the data, it dutifully removes it

nevir:
I don't think we should explicitly ignore _typename due to the union case,
though... Hmm

finnigantime:
@nevir Yeah I think it's the latter. In this case the ID has not changed, we're persisting a small update to the existing entity.

nevir:
[...]

PR #413, opened 2019-04-05
add warning around missing __typename in fragment payload

finnigantime:
✔ Write a summary

Add warning around missing __typename in fragment payload. See #412. This is my attempt to add a warning around this condition. Note that the warning is only emitted if verbose logging is enabled so I'm not sure how helpful it'll be.

✔ Tests!

N/A

✔ Bump the version if needed

N/A

### DIFF
--- a/src/context/CacheContext.ts
+++ b/src/context/CacheContext.ts
@@ -180,7 +180,8 @@ export class CacheContext {
   readonly tracer: Tracer;
 
   /** Whether __typename should be injected into nodes in queries. */
-  private readonly _addTypename: boolean;
+  readonly addTypename: boolean;
+
   /** All currently known & processed GraphQL documents. */
   private readonly _queryInfoMap = new Map<string, QueryInfo>();
   /** All currently known & parsed queries, for identity mapping. */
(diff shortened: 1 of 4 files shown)


src/context/CacheContext.ts
```diff
@@ -204,6 +204,14 @@ export class CacheContext {
     this._addTypename = config.addTypename || false;
   }
 
+  /**
+   * Returns a boolean indicating whether this CacheContext has been
+   * configured to add __typename when transforming documents.
+   */
+  getAddTypename() {
+    return this._addTypename;
+  }
```
nevir: Rather than converting this into a getter, let's make the property public

src/operations/SnapshotEditor.ts
```diff
@@ -223,6 +223,13 @@ export class SnapshotEditor {
         if (payload && payloadName in payload) {
           warnings.push(`Encountered undefined at ${[...prefixPath, ...path].join('.')}. Treating as null`);
         }
+
+        if (this._context.getAddTypename() && payloadName === '__typename') {
+          const existingTypenameValue = deepGet(this._getNodeData(containerId), path);
```
nevir: You can skip this check - it's even more of a problem for cases where typename is missing (for some consumers of hermes, __typename is part of object identity)

finnigantime: done. For unit testing, how to test? I didn't see an existing example checking tracer warnings. I see tests for catching InvalidPayloadError. By the way, should this be an InvalidPayloadError instead of a warning? That would be more visible to consumers.

nevir: Yeah, failing hard here would be 👍 

---

re: unit testing the tracer, we typically fall back to testing the console output, rather than the tracer itself (since that's more stable, and these messages are likely to change). But it's also not a very proven pattern; injecting the tracer might be better.

Example test: https://github.com/convoyinc/apollo-cache-hermes/blob/master/test/unit/Cache/transactions.ts#L53-L62

PR #439, opened 2020-06-25
Adds ability to migrate parametrized field by copying from same path

manuelconvoy:
✔ Write a summary

Adds ability to migrate parametrized fields by copying the data from same path, currently we only support migrating parametrized fields to a default value

✔ Tests!

Added a test for this particular scenario and we already have good coverage in this code path

✔ Bump the version if needed

* Adding a new, backwards-compatible, feature?  Sweet!
  Please increment the Y version in package.json.

### DIFF
--- a/src/operations/migrate.ts
+++ b/src/operations/migrate.ts
@@ -84,6 +84,9 @@ function migrateEntity(
           const copyFromNode = allNodes[copyFromFieldId];
           if (copyFromNode) {
             newData = copyFromNode.data;
+          } else {
+            // If copyFrom doesn't exist added so we can retrieve it on read
+            nodesToAdd[copyFromFieldId] = new ParameterizedValueSnapshot(newData);
           }
         }
         const newNode = new ParameterizedValueSnapshot(newData);
--- a/src/operations/migrate.ts
+++ b/src/operations/migrate.ts
@@ -28,6 +28,10 @@ export type ParameterizedMigrationEntry = {
   path: PathPart[],
   args: JsonObject | undefined,
   defaultReturn: any,
+  copyFrom?: {
+    path: PathPart[],
+    args: JsonObject | undefined,
+  },
 };
 export type ParameterizedMigrations = {
   // typename is the typename of the container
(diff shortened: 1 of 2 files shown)


src/operations/migrate.ts
```diff
@@ -72,7 +77,16 @@ function migrateEntity(
       // create a parameterized value snapshot if container doesn't know of the
       // parameterized field we expect
       if (!snapshot.outbound || !snapshot.outbound.find(s =>  s.id === fieldId)) {
-        const newNode = new ParameterizedValueSnapshot(parameterized.defaultReturn);
+        let newData = parameterized.defaultReturn;
+        if (allNodes && parameterized.copyFrom) {
```
jamesreggio, line 81: Where does `allNodes` come from / get set?

jamesreggio: Nevermind, I see it now below.

test/unit/operations/migrate/entitiesWithParameterizedFields.ts
```diff
@@ -293,4 +293,75 @@ describe(`operations.migrate`, () => {
     jestExpect(_.get(result, ['viewer', 'friends'])).toEqual([]);
   });
 
+  it(`can copy from path`, () => {
```
jamesreggio, line 296: Can you add a test where `copyFrom.args` is `undefined`? I want to ensure that the non-parameterized codepath works properly, since there are potentially some subtle differences.

src/operations/migrate.ts
```diff
@@ -72,7 +77,16 @@ function migrateEntity(
...
+          const copyFromFieldId = nodeIdForParameterizedValue(id, path, args);
+          const copyFromNode = allNodes[copyFromFieldId];
+          if (copyFromNode) {
+            newData = copyFromNode.data;
+          }
+        }
+        const newNode = new ParameterizedValueSnapshot(newData);
         nodesToAdd[fieldId] = newNode;
 
         // update the reference for the new node in the container
```
jamesreggio, line 92: Additional node references may be necessary if the data we're copying is itself an entity. I think @nevir could give a quick answer, though.

test/unit/operations/migrate/entitiesWithParameterizedFields.ts
manuelconvoy: Test added

PR #443, opened 2020-08-01
Parameterized entities with multi dimensional arrays

jdeff:
Hermes is not able to resolve parameterized nodes that are nested in a multi-dimensional array. Writes to the cache in this scenario is correct, but on reads, we were not calculating the path of the parameterized nodes correctly. This happens because multi-dimensional arrays were not considered when [traversing children](https://github.com/convoyinc/apollo-cache-hermes/blob/3d44bd1c01ab683531a0ab2fd3599571ba715eb3/src/operations/read.ts#L176). This change looks for the GraphQL objects at any depth.

### DIFF
--- a/src/operations/read.ts
+++ b/src/operations/read.ts
@@ -1,6 +1,6 @@
 import { CacheContext } from '../context';
 import { GraphSnapshot } from '../GraphSnapshot';
-import { ParsedQuery } from '../ParsedQueryNode';
+import { ParsedQuery, ParsedQueryNode } from '../ParsedQueryNode';
 import { JsonObject, JsonValue, PathPart } from '../primitive';
 import { NodeId, OperationInstance, RawOperation, StaticNodeId } from '../schema';
 import { isNil, isObject, walkOperation, deepGet } from '../util';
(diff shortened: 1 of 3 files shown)


src/operations/read.ts
```diff
@@ -172,19 +172,13 @@ export function _walkAndOverlayDynamicValues(
...
-          child = _wrapValue(child, context);
-          queue.push(new OverlayWalkNode(child as JsonObject, nextContainerId, node.children, nextPath));
-        }
+      if (_shouldWalkChildren(child, node)) {
+        child = _recursivelyWrapValue(child, context);
+        _expandValue(child, nextPath)
+          .map(c => new OverlayWalkNode(c.value, nextContainerId, node.children || {}, c.path))
+          .forEach(n => queue.push(n));
+      } else {
+        child = _wrapValue(child, context);
```
jdeff, line 181: hmm. I think I'm missing a push to the queue here. I wonder what missing unit test we have ...

jdeff, line 181: Hmmm. Look at this diff again, this `_wrapValue` doesn't appear to be in the original flow. I bet this isn't needed.

jdeff:
@nevir do you know where I can find some instructions on running the perf tests? Are these tests that I will need to write?

jdeff:
> However, you may want to double check perf, as this is a super hot path. The map/reduce calls may cause a fair bit overhead (vs iterating more directly). 
>
> Anecdotally, we had to drop a bunch of map and forEach calls in the past (using for (…) { syntax) to address serveral hot paths

I went ahead and changed the map/reduce calls to `for` loops.

PR #444, opened 2020-08-04
Update apollo client to v3 and TypeScript to latest

jdeff:
The main purpose of this change is to add support for the v3 apollo client. The biggest change affecting Hermes in version 3 is the consolidation of the apollo packages into one `@apollo/client`package. Some utilities from the deprecated `apollo-utilities` package have been removed, and the rest are now accessible via `@apollo/client/utilities`. For the utility methods that were removed, I copied the original implementations from `apollo-utilities` into Hermes. The `isEqual` utility is an exception, as the original implementation was just an alias to `@wry/equality`. For that utility, I just added a dependency to that package. 

There are a couple of changes in particular that I would like careful review:
- ApolloQueryable now *extends* ApolloCache instead of just implementing its interface. The reason I did this was because of all the missing non-abstract methods that exist now on ApolloCache (i.e. `gc`, `identify`, etc). These methods all have simple, almost no-op, implementations and it didn't seem right to copy them all into ApolloQueryable.
- I've updated Typescript to `@latest`, but I wonder if that is too far. It's currently at 3.9. I could look to use something lower if there are things in the `.d.ts` files that aren't compatible with older versions.
- The following methods have all been copied from `apollo-utilities` and I put them in places I thought were logical, but I am not super familiar with the project.
  - `valueFromNode`
  - `defaultValueFromVariable`
  - `variablesInOperation`
  - `getOperationDefinitionOrDie`

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,12 +37,12 @@
     "@types/chai": "^4.0.10",
     "@types/chai-as-promised": "^7.1.0",
     "@types/deep-freeze-strict": "^1.1.0",
-    "@types/graphql": "^0.12.4",
     "@types/lodash": "^4.14.102",
     "@types/lodash.findindex": "^4.6.3",
     "@types/lodash.isequal": "^4.3.3",
     "@types/lodash.set": "^4.3.3",
-    "@types/node": "^8.9.4",
+    "@types/node": "^14.0.27",
+    "@types/react": "^16.9.44",
     "audit-ci": "^2.4.0",
     "chai": "^4.1.2",
     "chai-as-promised": "^7.1.1",
@@ -53,23 +53,29 @@
     "eslint-plugin-import": "^2.8.0",
     "eslint-plugin-jest": "^22.0.0",
     "eslint-plugin-typescript": "^0.12.0",
-    "graphql": "^0.13.1",
-    "graphql-tag": "^2.6.0",
+    "graphql": "^15.3.0",
+    "graphql-tag": "^2.11.0",
     "jest": "^23.0.0",
     "jest-junit": "^5.0.0",
     "lodash": "^4.17.5",
+    "react": "^16.13.1",
     "rimraf": "^2.6.2",
-    "typescript": "~2.8.0",
+    "typescript": "^3.9.7",
     "typescript-eslint-parser": "^20.0.0"
   },
   "dependencies": {
-    "apollo-cache": "^1.0.0",
-    "apollo-utilities": "^1.0.0",
+    "@apollo/client": "^3.1.2",
+    "@wry/equality": "^0.2.0",
     "deep-freeze-strict": "^1.0.0",
     "lodash.findindex": "^4.0.0",
     "lodash.isequal": "^4.5.0",
     "lodash.set": "^4.0.0",
     "make-error": "^1.3.0",
-    "tslib": "^1.9.0"
+    "subscriptions-transport-ws": "^0.9.17",
+    "ts-invariant": "^0.4.4",
+    "tslib": "^2.0.0"
+  },
+  "peerDependencies": {
+    "@apollo/client": "^3.1.2"
   }
 }
--- a/src/apollo/Queryable.ts
+++ b/src/apollo/Queryable.ts
@@ -1,5 +1,5 @@
-import { Cache, DataProxy } from 'apollo-cache';
-import { removeDirectivesFromDocument } from 'apollo-utilities';
+import { ApolloCache, Cache, DataProxy, Reference, makeReference } from '@apollo/client';
+import { removeDirectivesFromDocument } from '@apollo/client/utilities';
 
 import { UnsatisfiedCacheError } from '../errors';
 import { JsonObject } from '../primitive';
@@ -11,7 +11,7 @@ import { buildRawOperationFromQuery, buildRawOperationFromFragment } from './uti
 /**
  * Apollo-specific interface to the cache.
  */
-export abstract class ApolloQueryable implements DataProxy {
+export abstract class ApolloQueryable<TSerialized> extends ApolloCache<TSerialized> {
   /** The underlying Hermes cache. */
   protected abstract _queryable: Queryable;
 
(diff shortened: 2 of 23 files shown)


package.json
```diff
@@ -53,23 +54,29 @@
...
     "eslint-plugin-jest": "^22.0.0",
     "eslint-plugin-typescript": "^0.12.0",
-    "graphql": "^0.13.1",
-    "graphql-tag": "^2.6.0",
+    "graphql": "^15.3.0",
+    "graphql-tag": "^2.11.0",
     "jest": "^23.0.0",
     "jest-junit": "^5.0.0",
     "lodash": "^4.17.5",
+    "react": "^16.13.1",
```
jdeff, line 62: `@apollo/client` has a dependency on `react` so TypeScript complains when it's unable to find it.

src/apollo/Queryable.ts
```diff
@@ -49,52 +49,50 @@ export abstract class ApolloQueryable implements DataProxy {
...
       [{ name: 'static' }],
       document
     )!;
   }
 
-  evict(options: Cache.EvictOptions): Cache.EvictionResult {
-    const rawOperation = buildRawOperationFromQuery(options.query, options.variables);
-    return this._queryable.evict(rawOperation);
+  evict(_options: Cache.EvictOptions): boolean { // eslint-disable-line class-methods-use-this
+    throw new Error(`evict() is not implemented in Hermes`);
```
jdeff, line 96: `Cache.EvictOptions` no longer includes a `query` so it's impossible to create a `RawOperation` from it. Since we don't really implement `evict` in any of the `Queryable` implementations, I just put the `throw` here.

src/apollo/Transaction.ts
```diff
@@ -1,5 +1,5 @@
-import { ApolloCache, Cache, Transaction } from 'apollo-cache';
-import { JsonValue } from 'apollo-utilities';
+import { Cache, Transaction } from '@apollo/client';
+import { Reference } from '@apollo/client/utilities';
```
jdeff, line 2: The apollo team moved away from `JsonValue` in favor of `Reference`. I'm not sure if any of our code is affected by this, though.

src/util/index.ts
```diff
@@ -2,4 +2,5 @@ export * from './ast';
 export * from './collection';
 export * from './primitive';
 export * from './references';
+export * from './store';
```
jdeff, line 5: Store may or may not be the right name here. Since I copied these functions from the `storeUtils.ts` file in `apollo-utilities`, I decided to keep a similar name.

test/env/base.ts
```diff
@@ -1,5 +1,5 @@
-import * as chai from 'chai';
-import * as chaiAsPromised from 'chai-as-promised';
+import chai from 'chai';
+import chaiAsPromised from 'chai-as-promised';
```
jdeff, line 2: This is due to enabling `esModuleInterop`.

src/apollo/Queryable.ts
```diff
@@ -49,52 +49,50 @@ export abstract class ApolloQueryable implements DataProxy {
...
       options.variables as any,
       options.fragmentName,
     );
     this._queryable.write(rawOperation, options.data as any);
-  }
-
-  writeData(): void { // eslint-disable-line class-methods-use-this
-    throw new Error(`writeData is not implemented by Hermes yet`);
+    return undefined;
   }
```
jamesreggio, line 81: I'm too lazy to look up what a `Reference` is, but if it's just an opaque ID (i.e., the result of `dataIdForObject`), we should support it to reduce future surprise.

Here's the description of what it is, but I'm not sure what you'd do with it: https://www.apollographql.com/docs/react/api/cache/InMemoryCache/#output-1

src/apollo/Queryable.ts
```diff
@@ -49,52 +49,50 @@ export abstract class ApolloQueryable implements DataProxy {
...
 
-  public transformForLink(document: DocumentNode): DocumentNode { // eslint-disable-line class-methods-use-this
+  transformForLink(document: DocumentNode): DocumentNode { // eslint-disable-line class-methods-use-this
     // @static directives are for the cache only.
     return removeDirectivesFromDocument(
       [{ name: 'static' }],
       document
     )!;
   }
```
jamesreggio, line 93: Is this still used? If not, we'll have a problem because the server will complain about `@static`.

test/unit/context/CacheContext/entityUpdaterOptimisticChange.ts
```diff
@@ -22,7 +22,8 @@ describe(`context.CacheContext`, () => {
         if (nextActive === prevActive) return;
 
         const userId = user ? user.id : previous.id;
-        const { activeUsers } = dataProxy.readQuery({ query: activeUsersQuery.document });
+        const { activeUsers } = dataProxy.readQuery({ query: activeUsersQuery.document }) as any;
```
jamesreggio, line 25: What is the new type of the `readQuery` result? Can you supply `<any>` as a generic type parameter instead to indicate you're not just smashing through things?

src/apollo/Queryable.ts
jdeff: yes. this is used in the `QueryManager` in `@apollo/client`.

src/apollo/Queryable.ts
jdeff: Yeah, I think we should look at doing that. I'll put up a follow-up PR to address this. In general, I think that we can improve the typing in this project.

jdeff:
> FYI: You can remove the dependency upon @types/graphql since they're now included in the graphql package as of v15.
Good point. I've removed this.

src/apollo/Queryable.ts
jdeff: Here's the type definition:

```
export interface Reference {
    readonly __ref: string;
}
```
It's an object with a `__ref` property. Still need to look into how it's used in the other caches.

test/unit/context/CacheContext/entityUpdaterOptimisticChange.ts
jdeff: It's just that it's possibly undefined and I can't de-structure it like this. I didn't feel like rewriting the test, though I probably should.

jdeff: actually, just tested this and it's no longer needed. I don't know what happened?

src/apollo/Queryable.ts
jdeff: Fixed.

PR #457, opened 2023-05-18
Attempt to fix publishing by switching to Artifactory

arama-convoy:
✔ Write a summary

Attempt to fix publishing by switching to Artifactory

✔ Tests!

Untestable without actually trying to publish


michaelcueno:
This is a public repo though right? I'm not sure if we should be publishing to our private artifactory. Maybe I'm missing something you know?

